PR #5, opened 2021-06-24
Az stg acct

danielle-casella-adams:
Adding a template for deploying to an az storage account.  Cleaning up some of the tf templates to consolidate their configuration.


hpractv (approved):
👍🏽

PR #7, opened 2021-07-08
ARCH-873 - Fixing some problems with the templates.

danielle-casella-adams:


workflow-templates/im-deploy-tf-manually-apply-to-prod-subscription.yml
```diff
@@ -161,6 +153,9 @@ jobs:
 
           echo "::set-output name=tf_plan_name::$terraformPlanName"
 
+      - name: Azure logout
+        run: az logout
```
JosephDSchwartz, line 157: The azure/login action recommends doing `az cache purge` and `az account clear` in addition to `az logout` (https://github.com/marketplace/actions/azure-login#az-logout-and-security-hardening). Is that something we want to do or nah? I don't know enough about azure and our runners to know if it's a good idea or not, but I thought I'd ask.

JosephDSchwartz (dismissed):
I have the one question, but it may be nothing so I'll go ahead and approve!

workflow-templates/im-deploy-tf-manually-apply-to-prod-subscription.yml
danielle-casella-adams: I didn't know that was a thing.  I will add it.

JosephDSchwartz (approved):
Your commit message legitimately made me laugh out loud 🤣

hpractv (approved):
👍🏽

PR #8, opened 2021-07-12
ARCH-873 - Templates for deploying code to an azure app service.

danielle-casella-adams:
- Include some fixes in other templates


hpractv (approved):
👍🏽

PR #9, opened 2021-07-14
ARCH-889 - Adding slot swapping & other template updates

danielle-casella-adams:
- Restrict the tf auto plan & comment & tf auto apply to dev on merge to when changes have been made to the tf files
- Fix the PAGERDUTY_SERVICE_IDS env var, it needs quotes around the secret
- Update add-or-update-json-properties action to v1.0.1 where there was a bug fix
- Introducing slot steps and a teams notification in the deploy az app service to dev/prod subscription.
- Update the checkmarx workflow so it only kicks off when there are changes to the src directory


hpractv (approved):
👍🏽

JosephDSchwartz (approved):

PR #14, opened 2021-07-27
ARCH-872 - Adding a template for the CI build.

danielle-casella-adams:
Also, rename one of the templates that still had 'copy' in the name.

This template was used copied to 1up and populated with the right values, so you can see it in action here: https://github.com/bc-swat/1UP/actions/runs/1072270799


hpractv (approved):
👍🏽

PR #15, opened 2021-07-27
ARCH-887 - starting/stopping/restarting apps

danielle-casella-adams:
Adding a workflow templates for starting/stopping/restarting azure app service and function apps.

If you want to see it in action:
https://github.com/bc-swat/1UP/actions/workflows/im-run-start-stop-restart-azure-app-service.yml


hpractv (approved):
👍🏽

PR #16, opened 2021-07-27
ARCH-888 - Adding a template for adding an az keyvault secret.

danielle-casella-adams:
If you want to see it in action:
https://github.com/bc-swat/1UP/actions/runs/1072825727


hpractv (approved):
👍🏽

PR #20, opened 2021-08-09
ARCH-745: Adding the build database ci workflow template

JosephDSchwartz:
This is my first attempt at creating one of these templates, so please judge it as harshly as possible! It's also a bit of a weird one because a lot of repos will only need half or less of what's in here.

### DIFF
--- a/workflow-templates/im-build-db-ci.yml
+++ b/workflow-templates/im-build-db-ci.yml
@@ -1,9 +1,9 @@
 name: Build Database CI
-on: 
+on:
   push:
-    branches: [ master ]
+    branches: [main]
   pull_request:
-    branches: [ master ]
+    branches: [main]
 env:
   DB_NAME: '' # TODO: Fill in your repo's database name
   DB_SERVER_NAME: 'localhost,1433'
@@ -48,13 +48,13 @@ jobs:
   lint-migration-files:
     runs-on: ubuntu-latest
     needs: check-for-snapshot-commit
-    if: ${{ needs.check-for-snapshot-commit.outputs.should-skip-remaining-workflow-jobs != 'true' }}
+    if: ${{ needs.check-for-snapshot-commit.outputs.should-skip-remaining-workflow-jobs == 'false' }}
     steps:
       - uses: actions/checkout@v2
 
       - uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '14' # TODO: Update this if necessary
 
       - id: migration-folder
         shell: pwsh
@@ -63,22 +63,21 @@ jobs:
         uses: im-open/tsql-lint-action@v1.0.0
         with:
           tsqllint-version: 1.11.0
-          path-to-sql-files: "Database/src/SqlDatabase/Migrations/${{ steps.migration-folder.outputs.folder }}" # TODO: Update this path to wherever the files to lint live
-          file-name-filter: "V9*.sql" # TODO: Update this if necessary
+          path-to-sql-files: 'Database/src/SqlDatabase/Migrations/${{ steps.migration-folder.outputs.folder }}' # TODO: Update this path to wherever the files to lint live
+          file-name-filter: 'V9*.sql' # TODO: Update this if necessary
           path-to-lint-config: ./.tsqllintrc # TODO: Update this with the path to your project's .tsllintrc file
-
 
   build-database:
     runs-on: [self-hosted, windows-2019-dev]
     needs: [check-for-snapshot-commit, lint-migration-files]
-    if: ${{ needs.check-for-snapshot-commit.outputs.should-skip-remaining-workflow-jobs != 'true' }}
+    if: ${{ needs.check-for-snapshot-commit.outputs.should-skip-remaining-workflow-jobs == 'false' }}
     steps:
       - uses: actions/checkout@v2
 
       - name: Install Flyway
         uses: im-open/setup-flyway@v1.0.0
         with:
-          version: 7.2.0
+          version: 7.2.0 # This version works with the current version of build-database-ci-action. Newer versions might, but they should be tested.
 
       - name: Build Database
         uses: im-open/build-database-ci-action@v1.0.1
@@ -96,7 +95,6 @@ jobs:
       # The steps that come next aren't going to be needed by every repo, so feel free to remove what is unnecessary for yours.
       #########################################################################################################################
 
-
       # The next four steps can be removed if your build doesn't need to create snapshots of database objects.
       # Otherwise they can be uncommented and the TODO's addressed.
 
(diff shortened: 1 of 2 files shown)


hpractv (approved):
👍🏽

workflow-templates/im-build-db-ci.yml
```diff
@@ -0,0 +1,186 @@
+name: Build Database CI
+on: 
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+env:
+  DB_NAME: '' # TODO: Fill in your repo's database name
+  DB_SERVER_NAME: 'localhost,1433'
+  DEFAULT_BRANCH: 'main'
```
danielle-casella-adams, line 10: We've got some mismatch here with the branch names.  The plan is to move everyone to main, not master, so I'd prefer to update line 4 and 6 with main and have it match line 10.

workflow-templates/im-build-db-ci.yml
```diff
@@ -0,0 +1,186 @@
...
+          {
+            Write-Host "Snapshot was not the last commit, continuing with the other jobs"
+            echo "::set-output name=skip::false"
+          }
+
+  # TODO: Remove this linting job if your project doesn't need/want it
+  lint-migration-files:
+    runs-on: ubuntu-latest
+    needs: check-for-snapshot-commit
+    if: ${{ needs.check-for-snapshot-commit.outputs.should-skip-remaining-workflow-jobs != 'true' }}
```
danielle-casella-adams, line 51: You did set the output to false above, what do you think about changing it to an affirmative just so there is less mental processing.
```suggestion
    if: ${{ needs.check-for-snapshot-commit.outputs.should-skip-remaining-workflow-jobs == 'false' }}
```

workflow-templates/im-build-db-ci.yml
```diff
@@ -0,0 +1,186 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
danielle-casella-adams, line 57: Do you think people will want to customize this?  It might be worth a # TODO:

workflow-templates/im-build-db-ci.yml
```diff
@@ -0,0 +1,186 @@
...
+          tsqllint-version: 1.11.0
+          path-to-sql-files: "Database/src/SqlDatabase/Migrations/${{ steps.migration-folder.outputs.folder }}" # TODO: Update this path to wherever the files to lint live
+          file-name-filter: "V9*.sql" # TODO: Update this if necessary
+          path-to-lint-config: ./.tsqllintrc # TODO: Update this with the path to your project's .tsllintrc file
+
+
+  build-database:
+    runs-on: [self-hosted, windows-2019-dev]
+    needs: [check-for-snapshot-commit, lint-migration-files]
+    if: ${{ needs.check-for-snapshot-commit.outputs.should-skip-remaining-workflow-jobs != 'true' }}
```
danielle-casella-adams, line 74: Same comment as above

workflow-templates/im-build-db-ci.yml
```diff
@@ -0,0 +1,186 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Flyway
+        uses: im-open/setup-flyway@v1.0.0
+        with:
+          version: 7.2.0
```
danielle-casella-adams, line 81: Do we want a TODO here for the version number?

workflow-templates/im-build-db-ci.yml
```diff
@@ -0,0 +1,186 @@
...
+
+      #     echo "::set-output name=json::$objectsAsJson"
+
+      # - name: Increment snapshots
+      #   uses: im-open/increment-database-object-snapshots@v1.0.0
+      #   with:
+      #     db-name: ${{ env.DB_NAME }}
+      #     instance-name: ${{ env.DB_SERVER_NAME }}
+      #     snapshot-path: ${{ env.SNAPSHOT_PATH }}
+      #     objects-to-increment: "${{ steps.changed-objects.outputs.json }}"
```
danielle-casella-adams, line 139: I think our standard is to do single quotes instead of double.  There are a number of places in this workflow file that using " instead of ' that you may want to fix.  @hpractv and I are both using YAML by Red Hat if that helps.    It may just be showing up like that though because they are in comments.
```suggestion
      #     objects-to-increment: '${{ steps.changed-objects.outputs.json }}'
```

workflow-templates/im-build-db-ci.yml
```diff
@@ -0,0 +1,186 @@
...
+      #   uses: im-open/sql-view-to-table-generator@v1.0.0
+      #   with:
+      #     schema-names: "dbo,CustomSchema" # TODO: Replace this with an actual list
+      #     db-name: ${{ env.DB_NAME }}
+      #     db-server: ${{ env.DB_SERVER_NAME }}
+      #     default-branch: ${{ env.DEFAULT_BRANCH }}
+      #     nuget-retrieval-url: "" # TODO: Replace this with the url for retrieving packages
+      #     publish-packages: true
+      #     nuget-publish-url: "" # TODO: Add the url for publishing packages
+      #     nuget-api-key: "${{ secrets.MY_GH_PACKAGES_ACCESS_TOKEN }}" # TODO: Update this with a real PAT
```
danielle-casella-adams, line 166: In the other workflows we've stayed away from adding placeholder or example values in the workflows.  We'd rather have it fail because it's missing than use a temp value that might work.
```suggestion
      #     nuget-api-key: '' # TODO: Update this with a real PAT
```

danielle-casella-adams (changes requested):
I think overall this looks good, there are some changes just based on convention and consistency with the other templates though.  

Also, have you been able to test this one out?  I've found the best way to do this is when the template is done to copy it into another project and just do the TODOs to make sure the syntax and everything is correct.

workflow-templates/im-build-db-ci.yml
JosephDSchwartz: This is the only one I know for sure works. With all of our windows vm problems I could only try a few others out, so there may be others, but I don't know if we want people changing this without knowing that they might run into problems. It's probably worth adding a comment saying something along those lines...

workflow-templates/im-build-db-ci.yml
JosephDSchwartz: I have no clue tbh. They might if the tsql linting stuff gets updated and needs a different version, but otherwise I imagine people will want to leave it alone. I think an `Update if necessary` TODO would cover that.

JosephDSchwartz:
Just so the answer to Danielle's question doesn't exist only in a private chat, I'll summarize here too.

I was able to test the workflow up to the point of building and running tests in the 1Up repo. However the steps after that are specific to the [ExtendHealth DB repo's build](https://github.com/im-practices/database). I did, however, test the individual scripts that are behind the actions used in the later steps. That's probably enough for now, but we'll need to be ready to debug problems that arise when we migrate that repo.

danielle-casella-adams (approved):
Thanks for making those changes!

PR #32, opened 2021-08-24
ARCH-999 - Force using github-hosted runners 

danielle-casella-adams:
Remove [] if there is only one label for the runs-on prop.


JosephDSchwartz (approved):
I thought using the `ubuntu-latest` tag would force the use of a github-hosted runner. I'm very sad that isn't the case 😢

PR #43, opened 2021-08-31
Condense dev/prod tf workflows

danielle-casella-adams:
This condenses the dev/prod terraform workflows into one.  I also made some of the new approval environment pattern changes in the app service/function deploys.


JosephDSchwartz (approved):
Wow, that was surprisingly few changes. Combining definitely seems like the right call.

PR #45, opened 2021-08-31
ARCH-1009 - Adding codenames to each of the workflows.

danielle-casella-adams:


JosephDSchwartz (approved):
I think ZestyFlamingo is my favorite, but it's hard to pick just one 😆

hpractv:
So some have `_v1` Are we only going to add ad version piece after we increment?

danielle-casella-adams:
InstaneHamster is my favorite.

danielle-casella-adams:
It looks like two were missing this _v1.  I'll push that change up.

hpractv (approved):
👍🏽

PR #46, opened 2021-08-31
ARCH-1010- set-vars job pattern

danielle-casella-adams:
- Implement the set-vars job pattern in the workflows that take in an environment.  
- Have each job print out which environment is being targeted as the first step.


hpractv (approved):
👍🏽

PR #48, opened 2021-09-07
ARCH-1018 - Replace increment logic with the git-version-lite action.

danielle-casella-adams:
Also fix the octostache template usage.


JosephDSchwartz (approved):
The action is way more complicated under the hood than what we were doing, but the usage of the action sure does clean things up nicely.

PR #55, opened 2021-09-14
ARCH-1043 - Changing the teams action to the new name (it's not really new)

danielle-casella-adams:


hpractv (dismissed):
The hampster's wheels are really churning.

hpractv (approved):
👍🏽

PR #58, opened 2021-09-16
ARCH-1052 - Adding a workflow for unlocking terraform state

danielle-casella-adams:


workflow-templates/im-run-unlock-tf-state.yml
```diff
@@ -0,0 +1,104 @@
...
+
+name: Unlock Terrafrom State
+on:
+  workflow_dispatch:
+    inputs:
+      lock-id:
+        description: The lock ID that needs to be removed.  Terraform provides this value (lock info.ID) in the message indicating there was an error acquiring the state lock.
+        required: true
+      environment:
+        description: The environment to deploy to - dev|qa|stage|prod|demo|uat
```
JosephDSchwartz, line 11: The environment to deploy to? Should this be something like "The environment with the lock" or something like that?

workflow-templates/im-run-unlock-tf-state.yml
```diff
@@ -0,0 +1,104 @@
...
+    ]
+
+jobs:
+  set-vars:
+    runs-on: ubuntu-latest # Force this to run on github-hosted runner by using a tag that does not exist on self-hosted runners
+
+    outputs:
+      ENVIRONMENT: ${{ steps.clean-env.outputs.mapped_input }} # To use this value: ${{ needs.set-vars.outputs.ENVIRONMENT }}
+      RESOURCE_GROUP: ${{ steps.rgrp.outputs.RESOURCE_GROUP }} # To use this value: ${{ needs.set-vars.outputs.RESOURCE_GROUP }}
+      STORAGE_ACCOUNT: ${{ steps.stg-acct.outputs.STORAGE_ACCOUNT }} # To use this value: ${{ needs.set-vars.outputs.STORAGE_ACCOUNT }}
```
JosephDSchwartz, line 41: Are these two leftovers from copy and paste? (I don't see them used below)

workflow-templates/im-run-unlock-tf-state.yml
```diff
@@ -0,0 +1,104 @@
...
+              "uat": ["u", "uat"],
+              "prod": ["prod", "production", "p"]
+            }
+          error_on_no_match: true
+          custom_error_message: 'The environment must be Dev, QA, Stage Demo, UAT or Prod'
+
+      - run: echo "The current environment is ${{ steps.clean-env.outputs.mapped_input }}"
+
+  tf-unlock-state:
+    needs: [set-vars, validate-tag-for-prod-deploys]
```
JosephDSchwartz, line 68: I don't see the `validate-tag-for-prod-deploys` job in this workflow.

workflow-templates/im-run-unlock-tf-state.yml
```diff
@@ -0,0 +1,104 @@
...
+    environment: ${{ needs.set-vars.outputs.ENVIRONMENT }}
+
+    defaults:
+      run:
+        shell: bash
+        working-directory: './${{ needs.set-vars.outputs.ENVIRONMENT }}' # TODO:  Verify this directory (./dev, ./qa, ./stage, ./prod, ./demo, ./uat) will be correct for your project
+
+    outputs:
+      tf_plan_name: ${{ steps.upload.outputs.tf_plan_name }}
+      maintenance_window_id: ${{ steps.open-window.outputs.maintenance-window-id }}
```
JosephDSchwartz, line 79: Do we need these outputs?

workflow-templates/im-run-unlock-tf-state.yml
```diff
@@ -0,0 +1,104 @@
...
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ env.GITHUB_REF }}
```
JosephDSchwartz, line 85: Does this need to be defined above? Or can it be removed?

danielle-casella-adams:
Man....I took a bunch of stuff out and the undid it because I realized I needed one section.  I guess I didn't go back and fix the rest of it.  Thank you for reviewing. :)

workflow-templates/im-run-unlock-tf-state.yml
danielle-casella-adams: I think it needs to be above.  Initially I thought I wouldn't need it but we need to do an init before running the unlock command so I think we do need it as an input to the workflow and we can set an env variable like the other workflows.

danielle-casella-adams:
@JosephDSchwartz this should be ready to look at again.

JosephDSchwartz (approved):
Looks all good to me now! Sorry about all of the questions, but I'm glad I caught some stuff 😃

danielle-casella-adams:
I don't mind the questions at all, they were all valid and helped prevent more re-work after it was merged!

PR #61, opened 2021-09-23
ARCH-1067 - verifying tags

danielle-casella-adams:
Adding an action to verify the tag is good and not mistyped for deploy workflow templates.  Adding the branch/tag/sha the user input to the echo statement that runs at the top of each job.  Refactory the nuget/npm workflows a bit (they probably still need more re-work to use our new deployment process).


workflow-templates/im-deploy-nuget-package.yml
```diff
@@ -73,7 +65,7 @@ jobs:
       #   run: dotnet pack --configuration Release
 
       # - name: Pack with version from env.Tag
-      #   run: dotnet pack --configuration Release -p:PackageVersion=${{ env.TAG }}
+      #   run: dotnet pack --configuration Release -p:PackageVersion=${{ steps.increment.outputs.VERSION }}
```
JosephDSchwartz, line 68: Should the name be updated to not reference env.Tag?

JosephDSchwartz (dismissed):
One tiny question, but otherwise I think this looks good. I'll approve in case you don't think a change is necessary!

workflow-templates/im-deploy-nuget-package.yml
danielle-casella-adams: yeah I'll do that.  I think I need to do some re-work here.  The nugets/npms were done before the deployment re-work so I think this could benefit from that.  It's still different though so I'm not sure how I want to handle it yet.  This one was very alternative heavy so there were lots of things to change.

danielle-casella-adams: I guess I misunderstood the question many times...you mean like 67?

workflow-templates/im-deploy-nuget-package.yml
```diff
@@ -73,7 +65,7 @@ jobs:
       #   run: dotnet pack --configuration Release
 
       # - name: Pack with version from env.Tag
```
danielle-casella-adams, line 67:
```suggestion
      # - name: Pack with version from git-version-lite
```

JosephDSchwartz (approved):

workflow-templates/im-deploy-nuget-package.yml
JosephDSchwartz: Sorry that my comment was confusing! But yes, I figured line 67 would be confusing since it _looks_ like a different value is being used on 68 (even though they're really the same thing).

PR #68, opened 2021-09-30
ARCH-1044 - attestor-approvals

danielle-casella-adams:
Adding the attestor approval job to function and app service deploys and on prem website deploy.


hpractv (approved):
👍🏽

PR #72, opened 2021-10-06
Variable by Scope Workflow Changes

hpractv:
- Added set-environment-variable-by-scope action
- Updated the octostache action

### DIFF
--- a/workflow-templates/im-deploy-az-function-manually.yml
+++ b/workflow-templates/im-deploy-az-function-manually.yml
@@ -177,6 +177,10 @@ jobs:
           #  - The name of the env (left side of @) (ex: Auth0_Domain).  This can be used in the later variable-substitution and octostache actions for replacement
           #  - The scope (right side of @) should be dev/qa/stage/uat/demo/prod. This action uses set-vars.ENVIRONMENT to match one of the scopes below and these are the values it will have.
           #  - The value for the environment should be included in '' after the colon. (ex: 'wtw-im-dev.auth0.com')
+          # Auth0_Domain@dev: 'wtw-im-dev.auth0.com'
+          # Auth0_Domain@qa: 'wtw-im-qa.auth0.com'
+          # Auth0_Domain@stage: 'wtw-im-stage.auth0.com'
+          # Auth0_Domain@prod: 'wtw-im-prod.auth0.com'
 
       # TODO:  If you need to do any Octopus variable substitution (i.e. replacing #{OCTO_PLACEHOLDER} in files) use the following action.  Delete if not using.
       # This action works well with the im-open/set-environment-variables-by-scope action. Environment variables set in that action will be utilized by this one.
--- a/workflow-templates/im-deploy-windows-service.yml
+++ b/workflow-templates/im-deploy-windows-service.yml
@@ -278,6 +278,10 @@ jobs:
           #  - The name of the env (left side of @) (ex: Auth0_Domain).  This can be used in the later variable-substitution and octostache actions for replacement
           #  - The scope (right side of @) should be dev/qa/stage/uat/demo/prod. This action uses set-vars.ENVIRONMENT to match one of the scopes below and these are the values it will have.
           #  - The value for the environment should be included in '' after the colon. (ex: 'wtw-im-dev.auth0.com')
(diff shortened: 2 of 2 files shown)


workflow-templates/im-deploy-az-app-service-manually.yml
```diff
@@ -45,70 +45,54 @@ jobs:
...
-              "prod": ["prod", "production", "p"] 
-            }
-          error_on_no_match: true
-          custom_error_message: 'The environment must be Dev, QA, Stage Demo, UAT or Prod' # TODO:  Update for the environments that are available
+          mapped_input@dev d development: 'dev'
+          mapped_input@qa q: 'qa'
+          mapped_input@stg s stage: 'stage'
+          mapped_input@o demo: 'demo'
+          mapped_input@u uat: 'uat'
+          mapped_input@prod p production: 'prod'
```
danielle-casella-adams, line 63: Why are these called mapped_input, can we not call it ENVIRONMENT so it's not ambiguous about what is being set?  

I also think this block should be combined with the one below.  Let's not call this action twice in one job.

workflow-templates/im-deploy-az-app-service-manually.yml
```diff
@@ -183,11 +167,11 @@ jobs:
...
       # TODO:  If you need to do any Octopus variable substitution (i.e. replacing #{OCTO_PLACEHOLDER} in files) use the following action.  Must be done once per file.  Delete if not using.
       # https://github.com/im-open/octostache-action
-      - uses: im-open/octostache-action@v1.0.0
+      - uses: im-open/octostache-action@v2.0.0
         with:
           variablesFile: '' # TODO: Add the ./path/file containing the variable substitutions to make, you will need to create this file  See the action for more details.
-          templateFile: '' # TODO:  Add the ./path/file to make the substitutions in
-          outputFile: '' # TODO:  Add the ./path/file to save the substitutions in (generally the same as templateFile above)
+          files-with-substitutions: '' # TODO:  Add the paths to the files to make substitutions in, e.g. ./path/file, ./path/another-file
+        # env: # TODO: Environment variables can be added that will also be used for substitution. These will override any variable in the variablesFile with the same name.
```
danielle-casella-adams, line 174:
```suggestion
        # env: # TODO: Environment variables can be added that will also be used for substitution (good for secrets). These will override any variable in the variablesFile with the same name.
```

workflow-templates/im-deploy-az-app-service-manually.yml
```diff
@@ -183,11 +167,11 @@ jobs:
 
       # TODO:  If you need to do any Octopus variable substitution (i.e. replacing #{OCTO_PLACEHOLDER} in files) use the following action.  Must be done once per file.  Delete if not using.
```
danielle-casella-adams, line 168: I think we need a new block im-open/set-environment-variables-by-scope block here with a # TODO to fill it out for anything need in appsettings.json replacement or octostatche replacement.  I can't put the suggestion in the right place in the file.  

The octostace/variable-substitution block comments should be updated to indicate they can use env values set in ^^ block within the job.

JosephDSchwartz: I went ahead and made the changes to the octostache action comment blocks. Let me know if the wording or anything isn't right.

workflow-templates/im-deploy-az-app-service-manually.yml
hpractv: I can change the output to `ENVIRONMENT`.  I had just left it to the same out as before so nothing obvious would break.

The real problem here is we're relying on user input for this step and the scope mapping doesn't line up the same for the possible key selection.

danielle-casella-adams: @hpractv  Did you see the second part of this comment about combining the actions in the set-vars job?

danielle-casella-adams: After trying it out I see why it wouldn't work so well to combine the two steps in set-vars job.

workflow-templates/im-deploy-az-app-service-manually.yml
danielle-casella-adams: I added the new block in the `deploy-code` section of the app service deploy.

PR #76, opened 2021-10-13
Combine app service and functions

danielle-casella-adams:


workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -172,42 +173,43 @@ jobs:
...
         env:
-          # TODO:  Add the values your app needs to be configured with (appsettings.json/octo var substitution/config changes). Envs that aren't used can be deleted.
-          #  - The name of the env (left side of @) (ex: Auth0_Domain).  This can be used in the later variable-substitution and octostache actions for replacement
-          #  - The scope (right side of @) should be dev/qa/stage/uat/demo/prod. This action uses set-vars.ENVIRONMENT to match one of the scopes below and these are the values it will have.
+          # TODO:  Add the values your app/function needs to be configured with (appsettings.json/local.settings.json/octo var substitution/config changes).  Delete envs that are not used.
+          #  - The name of the env (left side of @) (ex: Auth0_Domain) should be added.  The env name can be used in the later variable-substitution and octostache actions for replacement.
+          #  - The scope (right side of @) should be dev/qa/stage/uat/demo/prod. This action uses the value of set-vars.ENVIRONMENT to match one of the scopes (envs) below.
```
JosephDSchwartz, line 185: This is sort of a "while we're in here" kind of thing, but do we want to mention you can do multiple environments in one declaration, e.g. `Auth0_Domain@dev qa: 'wtw-im-dev-auth0.com'`?

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -231,25 +233,34 @@ jobs:
       # ARM_SUBSCRIPTION_ID is an env-level secret
       # - name: Create a deployment slot
       #   run: |
-      #     az webapp deployment slot create \
+      #     az ${{ env.APP_TYPE }} deployment slot create \
```
JosephDSchwartz, line 236: Would a name like `AZ_APP_TYPE` make it a little clearer what this env is for? I had to scroll all the way here before I knew for sure what it was being used to represent. The values were a clue, but I wasn't totally positive. This may just be me though.

workflow-templates/im-deploy-az-app-manually.yml
danielle-casella-adams: Yeah I can do that, stage is actually a perfect example because demo uses it as well.

JosephDSchwartz (approved):

PR #77, opened 2021-10-13
ARCH-1087 - Multiple deployables

danielle-casella-adams:
- Make sure the ref is explicitly set where possible for update deployment board
- Add a TODO for the deployable-type on update deployment board where it makes sense
- Increment version of update-deployment-board and cleanup-deployment-board
- Adding instructions at the top of the workflow indicating the changes that need to be made around multi-item deployments.
- Adding the workflow for deploying multiple items.


hpractv (approved):
👍🏽

PR #81, opened 2021-10-20
Adding a template for running a terraform import

JosephDSchwartz:


workflow-templates/im-run-tf-import.yml
```diff
@@ -0,0 +1,119 @@
...
+        description: The environment the import should be done in - dev|qa|stage|prod|demo|uat. # TODO:  Update for the environments that are available
+        required: true
+      branch-tag-sha:
+        description: The branch, tag or sha of the terraform that has the configuration for the resource(s) to import.
+        required: true
+      tf-resource-name:
+        description: The name of the resource to import.
+        requierd: false
+      tf-object-id:
+        description: The id of the object to import.
```
danielle-casella-adams, line 20: Do you think you could give examples of these?  I can see this being a point of confusion for what exactly needs to be added to both inputs.

workflow-templates/im-run-tf-import.yml
```diff
@@ -0,0 +1,119 @@
...
+      # Allows pulling modules from the repo instead of artifactory
+      - name: Setup SSH Keys and known_hosts
+        uses: im-open/setup-deploy-keys@1.0.1
+        with:
+          deploy-key-info: ${{ env.SSH_DEPLOY_KEY_INFO }}
+
+      - name: AZ Login
+        uses: azure/login@v1
+        with:
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
```
danielle-casella-adams, line 101: I don't think we need the az login because we are setting the ARM variables above....it would be nice if we could try it out first to see though.

workflow-templates/im-run-tf-import.yml
JosephDSchwartz: I think I was trying too hard to match the Octopus step template to be honest. It was very confusing and I have no clue why things were named the way they were. I've updated the names to more closely match terraform's documentation and also added what I hope are helpful examples.

danielle-casella-adams: I think those examples will really help.  It looks awesome!

PR #91, opened 2021-11-05
Small fixes from the first nuget migration

JosephDSchwartz:


hpractv (changes requested):
Do we need to increment the version?

hpractv (approved):
👍🏽

PR #92, opened 2021-11-11
Updating the build db workflow to run on linux

JosephDSchwartz:
For now there is a note that to run on windows you need to use integrated security. I plan on updating the Windows SQL Server installation on the runners to have the same account login creds as the Ubuntu ones so that is no longer necessary. Once that's done I'll update this template again.


hpractv (approved):
👍🏽

PR #98, opened 2021-11-23
ARCH-1108 - Updating the version numbers after standardizing a number of actions

danielle-casella-adams:


hpractv (approved):
When we get a chance, it would be nice to teach the increment version to add a major version only tag, "@v1".  Alot of these workflows would be fine and it would cut down on a lot of version churn.

PR #112, opened 2021-12-08
Bug fix: unzip files account access

hpractv:
Added `chmod` statement to allow the workflow user to update/edit files for configuration from the package zip.


izep (approved):

danielle-casella-adams (commented):
Was this to address an EACCCESS error?

hpractv:
> Was this to address an EACCCESS error?

No, it's just an access/edit thing.  Sometimes when you unpack a zip file from an untrusted source, you must explicitly set read/write/execute permissions on the new directory.  Mark, et. al, was getting access errors because the permissions weren't set.

danielle-casella-adams:
Ken had something similar and we solved it slightly differently.  The error he was getting started with EACCESS

PR #113, opened 2021-12-09
ARCH-1117: Set Vars Usage Link

hpractv:
- Added a link to the usage information for set-variables-by-scope action﻿


danielle-casella-adams (commented):
These are just comments so we don't need to increment the version of the workflow.

danielle-casella-adams (approved):

PR #116, opened 2021-12-17
Point the db build actions to version 3.0

JosephDSchwartz:


workflow-templates/im-build-db-ci.yml
```diff
@@ -134,20 +134,29 @@ jobs:
...
+          # incremental: false # TODO: This is the default value in the action. Delete this if you don't need to change it.
+          # create-database-file: '' # TODO: Add the path to the create database file (often called Initialization.sql)
+          # run-tests: true
+          # test-files-path: '' # TODO: Add the path to the directory with your test files
+          # test-timeout: '' # TODO: Add the time, in seconds, to wait for all the tests to run before timing out. The default is 300. Delete if 300 is fine or if you aren't running tests.
+          # drop-db-after-build: false # TODO: This is the default value in the action. Delete this if you don't need to change it.
+          # should-validate-migrations: false  # TODO: This is the default value in the action. Delete this if you don't need to change it.
+          # seed-data: true
+          # seed-data-files-path: '' # TODO: Add the path to the directory with your seed data files
+          # managed-schemas: dbo  # TODO: This is the default value in the action. If you need to change it add commas between the schemas (i.e. dbo,OtherSchema), otherwise delete this line.
```
danielle-casella-adams, line 159: What if instead of listing out every arg they could use we put in the most common items and then add a more detailed todo saying something like by default this runs on localhost port 1433 and will run tests and do seed data, etc.  Check out the action (with link) for more details on configuring db setup?

JosephDSchwartz: That sounds like a good idea! I think that would be less overwhelming for sure.

JosephDSchwartz: Ok I took a stab at those changes. Let me know what you think!

danielle-casella-adams (approved):
I like the changes!

PR #118, opened 2021-12-21
ARCH-814 - Adding a package.json so this can be restored via npm.

danielle-casella-adams:


JosephDSchwartz (approved):
The idea behind this seems pretty cool, I'm excited to hear if it works!

PR #119, opened 2021-12-21
ARCH-814 - Adding an inventory file for start-project integration.

danielle-casella-adams:


start-project-template-inventory.json
```diff
@@ -0,0 +1,64 @@
+{
+    "bff-with-react": {
+        "coreTemplates": [ 
+            { "name": "im-build-dotnet-ci.yml", "canHaveMultiples": false},
+            { "name": "im-deploy-az-app-manually.yml", "canHaveMultiples": true},
+            { "name": "im-deploy-multiple-items-at-once.yml", "canHaveMultiples": false},
```
JosephDSchwartz, line 6: This question will likely be answered when I see the start-project side of things, but I'm guessing having these two be "core" templates means users will have to remove one if they don't need it? I don't think that's necessarily a bad thing, I'm just wondering.

start-project-template-inventory.json
```diff
@@ -0,0 +1,64 @@
...
+            { "name": "im-deploy-cleanup-automated-board.yml", "canHaveMultiples": false},
+            { "name": "im-test-checkmarx-cli-sast-scan.yml", "canHaveMultiples": false}
+
+        ],
+        "optionalTemplates": [
```
JosephDSchwartz, line 11: Do users get to select which of these they want during the yeoman generation process?

JosephDSchwartz (approved):
I asked some questions just because I'm curious about how the process is going to work, but it looks great!

workflow-templates/im-test-cypress.yml
```diff
@@ -13,6 +13,7 @@ name: Run Cypress Tests
 #    "precy:ci": "rm -rf mochawesome-report/",
 # 4. Add the following script that merges all test json files and generates an html report.  Your name may vary based on your project conventions.  If the output file or results dir are changed, update the env values in the workflow.
 #    "cy:report": "mochawesome-merge mochawesome-report/*.json > merged-results.json && marge --reportDir cypress-results merged-results.json",
+#                                                                                        ^^ marge is not a typo, it's a command for the "mochawesome report generator"
```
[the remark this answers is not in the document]
danielle-casella-adams: I've probably answered that question like 20 times already. :)

start-project-template-inventory.json
danielle-casella-adams: The core templates will get put in their .github/workflows folder without any input.  The ones that are optional will only be included if they say yes, include those.  The core/optional follow what was laid out in git-er-done on the [recommended workflows](https://github.com/im-practices/git-er-done/blob/main/actions/recommended-templates.md#terraform-core-templates) page.

In this case the im-deploy-multiple-items-at-once.yml will probably be removed by most teams because they don't want to deploy all of their things at the same time.  That one is a bit unusual but it's kind of a "core" thing for people with that scenario.

start-project-template-inventory.json
danielle-casella-adams: They will get the core items by default and they are asked if they want the optional ones which they will get all or nothing from that group.  I'm probably going to do another PR in this repo that gives a better summary at the top of each workflow (that matches the git-er-done description) and also include what it's a core/optional template for.  That way when it gets pulled over in start-project, the user will have a better idea of what they can delete and what they should keep.

README.md
```diff
@@ -60,8 +62,19 @@ The templates in this repository fall into one of the following categories:
...
     - [ ] For any secrets, document whether it is an org, repo or environment level secret
       
+## start-project Integration
+This repository contains two files that enable integration with `start-project`:
+- [package.json]
+  - This allows the repository to be restored via npm
+- [start-project-template-inventory.json]
+  - This file tells `start-project` which workflow templates should be added to different project types
+
+When `start-project` runs, it restores this repository and examines the inventory file to determine which Workflows should be included.  The generator should include the latest version of the Workflow.
```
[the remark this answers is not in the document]
danielle-casella-adams: I usually will capitalize it when I'm speaking about the github concept of a thing but I don't always.

PR #120, opened 2021-12-23
ARCH-814 - Adding comments about the workflow

danielle-casella-adams:
Adding some standard comments to the top of each workflow indicating the purpose, frequency, projects to use it with and prerequisites.


workflow-templates/im-build-db-ci.yml
```diff
@@ -1,21 +1,28 @@
...
+#    There are TODO's throughout this template that let you know when something isn't necessary if you don't want snapshots and/or backups created.
+#             
+#    Creating mock sql objects is a bit less rare of a need, but only if your database is a sectioned off part of the legacy ExtendHealth monolith. Again,
+#    there will be TODO's showing you where some items can be removed if you don't have any dependencies and/or don't need to create mock objects to be
+#    consumed by others.
 #
-#       The need for creating snapshots and backups in this workflow is not a very common need. They are here in case you need them, but likely you won't.
-#       There are TODO's throughout this template that let you know when something isn't necessary if you don't want snapshots and/or backups created.
+# Frequency: 
+#    - This workflow should only be used once per repository
```
JosephDSchwartz, line 21: Unless they have multiple databases in a repo and want to build them separately I guess. I can't imagine anyone doing that so this is probably fine. 

* Edit: I think this probably goes for all of the build workflows. But again, since that seems unlikely to be needed, if it ever is the team who needs it can figure it out (probably by asking us).

JosephDSchwartz (approved):
I think everything looks good. When things are super repetitive like this though my brain starts to skip things though, so there could be a small typo or something. But all of these changes are descriptive comments so it doesn't matter. :shipit:

workflow-templates/im-build-db-ci.yml
danielle-casella-adams: There's a bunch of workflows I think that applies to, it doesn't seem likely that it would happen but it definitely could.

PR #122, opened 2022-01-21
ARCH-894 & ARCH-893 Add notes to the workflows to point people at the actions

JosephDSchwartz:
Danielle and I talked and we decided that, since most teams won't be using these actions, it would be a lot of noise to implement them in the workflow templates. A lot of these templates are already huge, so anything niche like these actions feel even more out of place than they would otherwise.

### DIFF
--- a/workflow-templates/im-deploy-az-app-manually.yml
+++ b/workflow-templates/im-deploy-az-app-manually.yml
@@ -1,17 +1,17 @@
 # Workflow Code: AmbitiousLizard_v23    DO NOT REMOVE
-# Purpose: 
-#    Gathers various stakeholder and attestor approvals, downloads artifacts from a release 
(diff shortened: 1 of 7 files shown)


hpractv (approved):
👍🏽  At least we don't have to increment all the versions on this one.

JosephDSchwartz:
I'm on the fence about incrementing the versions. I think we've done it for comment-only changes before, but I don't know if we care about tracking if people have these particular changes. I'm leaning towards no, but I'm willing to hear arguments for yes.

mike-schenk:
> I'm on the fence about incrementing the versions. I think we've done it for comment-only changes before, but I don't know if we care about tracking if people have these particular changes. I'm leaning towards no, but I'm willing to hear arguments for yes.

FWIW, I agree with "no"

danielle-casella-adams (approved):

PR #124, opened 2022-01-26
ARCH-1146 - Remove central logging

danielle-casella-adams:
Removing central logging ssh key from each of the workflows since it is now deprecated.


JosephDSchwartz (commented):
`im-build-tf-auto-plan-and-comment-on-prs.yml`'s version number wasn't updated like the rest of the workflows. Was that intentional?

danielle-casella-adams:
> `im-build-tf-auto-plan-and-comment-on-prs.yml`'s version number wasn't updated like the rest of the workflows. Was that intentional?

It was not on purpose.  I had the change locally but I guess I hadn't saved before pushing the rest of the file up.

JosephDSchwartz (approved):

PR #128, opened 2022-02-09
ARCH-1092: Add credential retrieval to on-prem database deploy workflow

JosephDSchwartz:


workflow-templates/im-deploy-on-prem-database.yml
```diff
@@ -146,6 +147,8 @@ jobs:
     needs: [set-vars, validate-tag-for-prod-deploys]
     runs-on: [self-hosted, ubuntu-20.04]
     environment: ${{ needs.set-vars.outputs.ENVIRONMENT }}
+    env:
+      VAULT_URL: https://vault.myvault.com:8200
```
danielle-casella-adams, line 151: That's not the actual url for the production vault is it?

related to that, we use the prod vault for storing dev/qa/stage/prod db creds right?

workflow-templates/im-deploy-on-prem-database.yml
```diff
@@ -177,6 +180,32 @@ jobs:
           DB_SERVER_NAME@stage demo: ''
           DB_SERVER_NAME@prod: ''
 
+      - name: AZ Login
+        uses: azure/login@v1
+        with:
+          # TODO: These the Service Principal these credentials are for needs access to Vault. See this git-er-done article for more information:
```
danielle-casella-adams, line 186: This might need some re-wording.

workflow-templates/im-deploy-on-prem-database.yml
```diff
@@ -177,6 +180,32 @@ jobs:
...
+          # https://github.com/im-practices/git-er-done/blob/main/actions/hashicorp-vault-integration.md#setup
+          creds: ${{ secrets.AZURE_LOGIN_CREDENTIALS }}
```
danielle-casella-adams, line 188: If these are the same JSON formatted creds that we use in our tf, should we call it AZURE_CREDENTIALS?  That's how the secret is referred to everywhere else.

workflow-templates/im-deploy-on-prem-database.yml
```diff
@@ -177,6 +180,32 @@ jobs:
...
+
+      - name: Import Secrets
+        id: vault-secrets
+        uses: hashicorp/vault-action@v2.3.1
+        with:
+          url: '${{ env.VAULT_URL }}'
+          token: '${{ steps.vault_token.outputs.client_token }}'
+          secrets: |
+            secret/data/ci/auth-creds username | SQL_AUTH_USERNAME ;
+            secret/data/ci/auth-creds password | SQL_AUTH_PASSWORD
```
danielle-casella-adams, line 207: I got lots of questions on the workflow where we use `marge`.  Since this looks not right I would suggest adding a comment here so people don't mess with that formatting because they think it's a typo.

hpractv (approved):
👍🏽

workflow-templates/im-deploy-on-prem-database.yml
JosephDSchwartz: ...did I have a stroke while I wrote this?!

workflow-templates/im-deploy-on-prem-database.yml
JosephDSchwartz: Good point. You already asked about that in a different review. A comment would probably help cut off problems/questions before they happen.

workflow-templates/im-deploy-on-prem-database.yml
```diff
@@ -170,28 +171,60 @@ jobs:
...
-          DB_NAME@stage demo: ''
+          DB_NAME@stage demo uat: ''
           DB_NAME@prod: ''
           DB_SERVER_NAME@dev: ''
           DB_SERVER_NAME@qa: ''
-          DB_SERVER_NAME@stage demo: ''
+          DB_SERVER_NAME@stage demo uat: ''
           DB_SERVER_NAME@prod: ''
+          VAULT_URL@dev qa: 'https://vault.dev.mktp.io:8200'
+          VAULT_URL@stage demo uat prod: 'https://vault.mktp.io:8200/'
```
danielle-casella-adams, line 181: What do you think about TODOs for both of these lines indicating the user needs to verify the vault url for their environments since it's unclear how people are using the two vault environments.

danielle-casella-adams (approved):
I think this looks great, just one thought about adding a TODO but you don't have to do that if you don't want to.

workflow-templates/im-deploy-on-prem-database.yml
JosephDSchwartz: That seems like a good idea. I was thinking the general TODO at the top of the step that says to fill out the values would be good enough, but since these look filled out already I think it would be good to add something specifically for them.

PR #131, opened 2022-02-18
ARCH-1171 - Give workflow_dispatch inputs a type

danielle-casella-adams:
- Where possible remove the set-vars job since setting the environment was a large part of that.  Consolidate jobs if there is a single job or only a few inputs that were set.


workflow-templates/im-deploy-az-database.yml
```diff
@@ -30,110 +30,78 @@ on:
...
-    steps:
-      - name: Checkout Repository
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-
-      - name: Verify Tag Exists
-        uses: im-open/verify-git-ref@v1.0.1
-        with:
-          branch-tag-sha: ${{ env.RELEASE_TAG }}
```
JosephDSchwartz, line 73: I honestly can't remember why we had this check in addition to the `is-tag-reachable-from-default-branch` check, but it seems like we're ok with deleting it now? Just double checking.

JosephDSchwartz (commented):
Kind of amazing how much stuff we can get rid of now that GitHub has introduced the simple concept of input types.

hpractv (dismissed):
That cleans up a lot busy work!

workflow-templates/im-deploy-az-database.yml
danielle-casella-adams: I actually think I got a little delete-happy and it should be there.  It's to ensure someone didn't fat-finger the tag before the workflow really gets started.

danielle-casella-adams: I put that back in and fixed it in a couple of other places where I got a little too excited to delete things. :)

JosephDSchwartz: Ah that's right, I remember now! Glad I double checked 😄

PR #135, opened 2022-03-16
ARCH-1190 - Update any usages of `env.ENVIRONMENT` in `environment:`

danielle-casella-adams:
Update any usages of `env.ENVIRONMENT` to `github.event.inputs.environment` when used to populate `environment:`.  

The env context is not available to jobs.id.environment. You can only use things from the github, needs, strategy, matrix or inputs contexts in that field. Which makes sense because before we would have been using an output from the set-vars job (the needs context). Using the env.ENVIRONMENT variable in other places where just fine where it had access to the env context so there was no naming clash which is initially what I thought might be happening.


hpractv (approved):
👍🏽

PR #138, opened 2022-03-31
ARCH-1205 - Updating the commitish behavior for create-release

danielle-casella-adams:
- Also comment out the excluded paths in the checkmarx workflow.  I didn't increment the version because it's not really changing functionality, it's just commenting out a thing most people aren't using.


hpractv (approved):
👍🏽

PR #142, opened 2022-04-27
- Fix the examine-triggers functionality.

danielle-casella-adams:
- Updating versions of actions for dependabot.
- Add a dependabot.yml to see if it will look at our workflow-templates folder.


.github/workflows/increment-version-and-kick-off-sync.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
       # See https://github.com/im-open/git-version for more details around how to increment major/minor/patch through commit messages
       - name: Increment the version
-        uses: im-open/git-version-lite@v2.0.6
+        uses: im-open/git-version-lite@v2.0.8
```
hpractv, line 20: Do we not put a worklfow version on this, because the teams don't really edit this?

hpractv (approved):
I just have a wondering question about the increment version workflow that may not be pertinent to deployment.

.github/workflows/increment-version-and-kick-off-sync.yml
danielle-casella-adams: Yeah this didn't have a version number because this is the one that belongs to the repo that it uses to version itself.

PR #143, opened 2022-05-16
Remove unused workflow check

bradyclifford:
This workflow step is now obsolete.

This breaks the workflow currently.


mike-schenk (commented):
I suppose, technically the version number at the top of the file ought to be incremented with this change.

mike-schenk:
Using #146 from a branch instead of this from a fork so the sync process will work.

PR #144, opened 2022-05-16
Update same comment with release version

bradyclifford:
We have seen QA getting confused with what version / tag they should be deploying when clearing a PR. But updating the same comment with the updated tag version after a commit, this resolved the problem.


mike-schenk (changes requested):
The version number in line 1 of this file needs to be incremented.

mike-schenk (approved):

PR #147, opened 2022-05-23
ARCH-1222: Add AsyncAPI document generation to the dotnet build workflow

JosephDSchwartz:


JosephDSchwartz:
Related to https://github.com/im-practices/cloud-events/pull/11

mike-schenk:
> Would it make sense to add this to a separate workflow instead. The CI build workflow is getting very large. I like what @stevematney did moving his Cypress tests out into their own workflow.

As a template, I think I like the workflow being all-inclusive. If a team wants to break up their own copy, that's fine.

Also, for now, we want generated AsyncAPI documents to be uploaded to release assets along with the deployable code, so it makes sense to create it in the same workflow where we package and upload deployables.

hpractv (approved):
👍🏽

PR #155, opened 2022-06-03
New build and test template to be used by start-project

mike-schenk:
Reworking start-project-template-inventory to coordinate with changes in https://github.com/im-practices/start-project/pull/183


workflow-templates/im-build-and-test-generic.yml
```diff
@@ -0,0 +1,328 @@
...
+            const continueWorkflow = prClosed && !prMerged ? false : true;
+            const doTagRelease = continueWorkflow && !prIsDraft ? true :  false;
+
+            core.exportVariable('IS_PRERELEASE', isPreRelease);
+            core.exportVariable('REF_TO_BUILD_AND_TAG', refToBuildAndTag);
+            core.exportVariable('CONTINUE_WORKFLOW', continueWorkflow);
+            core.exportVariable('CREATE_RELEASE', doTagRelease);
+
+      # NOTE: This workflow uses one version tag with no prefix for all the deployables in the repository.
+      #       If you want each deployable to have it's own version tag prefix, this workflow is not for you.
```
JosephDSchwartz, line 74: Do you think it makes sense to put this higher up so it's easier to see? And would it make sense to point people to some other template they can use if they have that use case? Or maybe give instruction on how to update this workflow, something along the lines of [this](https://github.com/im-practices/.github/blob/main/workflow-templates/im-build-dotnet-ci.yml#L375)?

Also, I think we've trained people to search for TODO's when implementing these templates, but not NOTE's, so it might make sense to make this a TODO if we leave it here. If we move it up to the top, it could probably go along with section that specifies what projects to use this template for.

workflow-templates/im-build-and-test-generic.yml
```diff
@@ -0,0 +1,328 @@
+# Workflow Code: MadLark_v1    DO NOT REMOVE
+# Purpose:
+#    Provide a framework for building and testing multiple project types in 
+#    this repository in parallel jobs in response to pull requests.
+#    Each build-test job should check out the code, build one project, run its
+#    associated tests and upload the packaged binaries as an artifact.
+#    After all the build-test jobs have completed, the create-release job
+#    will upload the built artifacts to a GitHub release asset.
```
JosephDSchwartz, line 8: Should this template only be implemented by the start-project process? I guess a team could manually implement it, but then some of the `(don't delete this comment)` comments won't make be applicable (if I'm understanding things correctly). Do we want to give further instruction along those lines?

mike-schenk: Because this template is specifically built to work with `start-project`, and `start-project` can add it to all the repos it touches, I think I'm going to actually remove it from this repo and add it to start-project itself instead.

JosephDSchwartz (approved):

PR #162, opened 2022-06-21
ARCH-1549 - Updating action versions after dependabot updates.

danielle-casella-adams:


.github/workflows/increment-version-and-kick-off-sync.yml
```diff
@@ -35,7 +35,7 @@ jobs:
 
       # See https://github.com/im-open/git-version for more details around how to increment major/minor/patch through commit messages
       - name: Increment the version
-        uses: im-open/git-version-lite@v2.0.8
+        uses: im-open/git-version-lite@v2.0.10
```
JosephDSchwartz, line 38: Should the version of this workflow file be updated like the others? I wondered if the changes in `git-version-lite` were insignificant enough to warrant it, but we updated `im-build-increment-version-on-merge` where that was the only change too, so I'm thinking maybe this one should be updated as well?

JosephDSchwartz (approved):

PR #167, opened 2022-08-25
Updating im-test-sam-yaml to use self-hosted runners.

stevematney:
Because the stewardship export is on by default, this must run on a self-hosted runner, otherwise the json upload will come from the public internet and will be blocked by Cosmos IP restrictions.

### DIFF
--- a/workflow-templates/im-test-sam-yaml.yml
+++ b/workflow-templates/im-test-sam-yaml.yml
@@ -1,4 +1,4 @@
-# Workflow Code: DigitalWerewolf_v2    DO NOT REMOVE
+# Workflow Code: DigitalWerewolf_v3    DO NOT REMOVE
 # Purpose:
 #    Runs the SAM YAML Validation action and exports the
 #    output JSON into the Stewardship Mapping document repository
--- a/workflow-templates/im-test-sam-yaml.yml
+++ b/workflow-templates/im-test-sam-yaml.yml
@@ -29,7 +29,7 @@ on:
 jobs:
   sam-yaml:
     name: Validate and Export SAM YAML
-    runs-on: [ubuntu-latest]
+    runs-on: [self-hosted, ubuntu-20.04]
 
     steps:
       - name: Checkout Repository


workflow-templates/im-test-sam-yaml.yml
```diff
@@ -29,7 +29,7 @@ on:
 jobs:
   sam-yaml:
     name: Validate and Export SAM YAML
-    runs-on: [ubuntu-latest]
+    runs-on: [self-hosted, ubuntu-20.04]
```
danielle-casella-adams, line 32: What's the thinking for changing this to our self-hosted runners?

stevematney: It's in the PR description. 😄
> Because the stewardship export is on by default, this must run on a self-hosted runner, otherwise the json upload will come from the public internet and will be blocked by Cosmos IP restrictions.

stevematney: You can see an example of this failure here: https://github.com/im-client/help-and-support/runs/8025578592?check_suite_focus=true#step:4:8

hpractv: It's true the firewall protection rules won't let machines from "off-prem" or unknown networks write to it now.  I recently had to add it to the cosmos account because of a policy that's going to be enforced mid Sept.

hpractv (dismissed):
👍🏽

mike-schenk:
@hpractv @danielle-casella-adams, so it sounds like, somebody (😉) needs to go through all the existing repos and update this workflow in them all.

JosephDSchwartz (changes requested):
Just one small detail we need to take care of: the workflow code version at the top of the `im-test-sam-yaml.yml` file needs to be incremented.

hpractv:
> Just one small detail we need to take care of: the workflow code version at the top of the `im-test-sam-yaml.yml` file needs to be incremented.

@JosephDSchwartz is right. I just realized it seconds before reading Joe's comments.

workflow-templates/im-test-sam-yaml.yml
danielle-casella-adams: @hpractv you might need to take those out or greenlight the current github ips so the existing workflows don't break until we get a chance to update it.

stevematney:
> the workflow code version at the top of the im-test-sam-yaml.yml file needs to be incremented.
Done!

JosephDSchwartz (approved):

PR #169, opened 2022-09-08
Update the run flyway command and add new instructions.

JosephDSchwartz:
The run-flyway-command action now allows for using an azure service principal or managed service identity to authenticate with the database. We're updating the action's version and adding some instructions around that.

### DIFF
--- a/workflow-templates/im-deploy-az-database.yml
+++ b/workflow-templates/im-deploy-az-database.yml
@@ -138,6 +138,7 @@ jobs:
         with:
           ref: ${{ env.RELEASE_TAG }}
 
+      # TODO: If your database uses Azure Active Directory Authentication then this step can be removed
       - name: AZ Login
         uses: azure/login@v1
         with:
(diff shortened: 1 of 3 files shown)


workflow-templates/im-deploy-az-database.yml
```diff
@@ -184,8 +185,11 @@ jobs:
           echo "SQL_USER=$sqlUser" >> $GITHUB_ENV
           echo "SQL_PASSWORD=$sqlPassword" >> $GITHUB_ENV
 
+      # TODO: If your database uses Azure Active Directory Authentication then you can login with a Service Principal
+      #       or Managed Service Identity instead of using the username and password of an account stored in Key Vault.
```
mike-schenk, line 189: Logging in with a managed service identity wouldn't be appropriate from our GitHub action runners. Only a service principal would make sense.

workflow-templates/im-run-flyway-repair.yml
```diff
@@ -129,8 +130,11 @@ jobs:
       #       database/static-creds/<role-name> username | SQL_USERNAME ;
       #       database/static-creds/<role-name> password | SQL_PASSWORD
 
+      # TODO: If your database uses Azure Active Directory Authentication then you can login with a Service Principal
+      #       or Managed Service Identity instead of using the username and password of an account stored in Key Vault.
```
mike-schenk, line 134: Logging in with a managed service identity wouldn't be appropriate from our GitHub action runners. Only a service principal would make sense.

mike-schenk (approved):

PR #171, opened 2022-09-22
ARCH-1601 - Multi-machine deploys

danielle-casella-adams:
Update iis site and windows service workflows for multiple machine deploys

The deploy jobs are what will be run multiple times via a matrix.  I was hoping to cut down as much as possible on the repetitive jobs like the octostache/variable substitution but I think that's just going to have to be run multiple times.  

We could potentially cut out the duplication if the deploy job did all the setup for octostache/variable substitution etc and then called the reusable workflow multiple times but you would have to [pass in all of the environment variables](https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations) that you set up for the action (which can be significant).  
>Any environment variables set in an env context defined at the workflow level in the caller workflow are not propagated to the called workflow.

This repetition seems better than that.  I'm also not super worried about this since it will be used with on-prem deploys which will hopefully modernize in the next year or so and go to azure.

The set-vars job was re-introduced so the deploy specific items could be set in that job and wouldn't have to be set for each deploy job.  It also gives us a chance to set the array of machines that we will use in the later deploy jobs to set up our matrices via the `needs` context.
https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
|Workflow key | Context|
|---|---|
|`jobs.<job_id>.strategy `|`github, needs, inputs`|

### DIFF
--- a/workflow-templates/im-deploy-windows-service.yml
+++ b/workflow-templates/im-deploy-windows-service.yml
@@ -219,6 +219,10 @@ jobs:
         # One job will spawn for each server listed in the array
         windows-server: ${{ needs.set-vars.outputs.WINDOWS_SERVERS }}
 
+    defaults:
+      run:
+        shell: pwsh
+
     env:
       PAGERDUTY_WINDOW_IN_MIN: 30 # TODO: Verify the length of your PD Maintenance Window
       PAGERDUTY_WINDOW_DESC: 'Deploying Code to ${{ github.event.inputs.environment }} from GitHub Actions' # TODO: Verify this PD Maintenance Window Description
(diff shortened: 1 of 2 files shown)


workflow-templates/im-deploy-windows-service.yml
```diff
@@ -260,43 +282,44 @@ jobs:
...
           # TODO: replace examples with actual substitutions
           # FedAuth.Domain: ${{ env.FedAuth_Domain }} # Example showing replacement at a nested node (left side) & using the env set in the set-environment-variables-by-scope step above (right side)
           # SecretValue: ${{ secrets.VALUE }} # Example showing replacement of a root level value (left side) with a secret (right side)
 
       - name: Zip the published app for faster deployment and uploads
-        working-directory: ${{ env.UNZIPPED_ASSET }}
-        run: zip -r ../${{ env.DEPLOY_ZIP }} .
+        run: |
+          # On a Windows build runner, the 'zip' commandline tool isn't available. Use PowerShell functions instead.
+          Compress-Archive -Path .\${{ env.UNZIPPED_ASSET }}\* -DestinationPath ${{ env.DEPLOY_ZIP }}
```
JosephDSchwartz, line 301: Is this job setting the default shell to powershell? It doesn't look like it is, but I could be blind.

danielle-casella-adams: I'm glad you found this!  I kept thinking I needed to take a look at the run statements and make sure it would work with everything.  I set the default on the other workflow but not this one.

JosephDSchwartz (approved):

PR #175, opened 2022-09-28
Tag the correct commit

mike-schenk:
This works around a current limitation in `git-version-lite`, which has a hard-coded condition to tag the head of the incoming branch whenever the workflow is triggered by a PR even if that PR is now merged.

I argue that `git-version-lite` probably should not push tags at all; probably we should use `create-release` to do it. But if it's going to, it needs to accept—maybe require—a committish parameter.

This PR preserves the existing behavior of creating only a tag and not a GitHub release. I could be convinced we should create releases here instead.


izep (approved):

hpractv (approved):
👍🏽

danielle-casella-adams (approved):

bradyclifford:
Is this change important enough that it needs to be updated in the existing workflows in our repos?

mike-schenk:
> Is this change important enough that it needs to be updated in the existing workflows in our repos?

It's not critically important _if_:

1. The team always uses merge commits to merge PRs (not "squash and merge" and not "rebase and merge").
2. The `main` branch protection rules require that the branch be up-to-date before merging every PR.

If both are true, the tag will still be on the "wrong" commit, but the _contents_ of the tagged commit will be identical to the contents of the merge commit that follows, so it isn't so bad.

If either of those is not true, it's very important to take this update. If 1 is not true, git-version-lite and tagging will fail on subsequent runs and deployment checks that require tags to be reachable from `main` will fail. If 2 is not true, the tag will not point to the correct merged contents leading to a potential regression when deployed.

PR #179, opened 2022-10-28
Use NEXT_VERSION_NO_PREFIX arg in nuget workflow

danielle-casella-adams:
Tweak nuget workflow to use NEXT_VERSION_NO_PREFIX instead of manipulating the NEXT_VERSION string


hpractv (approved):
👍🏽

PR #193, opened 2023-01-20
SRE-241 - K6 Operator workflow updates

jsclifford:
* Updated to latest actions and annotation improvements.


cbchadb35 (approved):

danielle-casella-adams (commented):
It doesn't look like the version has been incremented on any of these workflows.  The first one may not need it since you are just changing a name, the other two do since you are changing versions and updating jobs/steps.

jsclifford:
Updated versions @danielle-casella-adams

danielle-casella-adams (approved):

PR #194, opened 2023-01-20
SRE-243 - Updated for failover regions deployments

jsclifford:
* Updated Workflows to handle failover region deploys
* Added Step Summary for tf plan.


workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -103,27 +110,71 @@ jobs:
...
           RESOURCE_GROUP@uat: ''
           RESOURCE_GROUP@prod: ''
+          RESOURCE_GROUP@prod-secondary: ''
+
+          # Primary resource group used to pull azure credentials and secrets from
+          PRIMARY_RESOURCE_GROUP@dev: ''
+          PRIMARY_RESOURCE_GROUP@qa: ''
+          PRIMARY_RESOURCE_GROUP@stage stage-secondary: ''
+          PRIMARY_RESOURCE_GROUP@demo: ''
+          PRIMARY_RESOURCE_GROUP@prod prod-secondary: ''
```
mike-schenk, line 128: Unless I'm misunderstanding something, credentials ad secrets are not pulled from resource groups.

From reading the rest of the code, I guess the comment should be changed to say something more along the lines of `# Resource group holding the state storage account and managed service identities`

mike-schenk: Also, now that there are two variables holding resource group names, it might be better to rename the existing `RESOURCE_GROUP` variable to something like `TARGET_RESOURCE_GROUP` or `DESTINATION_RESOURCE_GROUP`

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -103,27 +110,71 @@ jobs:
...
           # This variable is used to deploy the app, swap slots, annotate app insights and send a teams notification
           AZ_APP_NAME@dev: ''
           AZ_APP_NAME@qa: ''
           AZ_APP_NAME@stage: ''
+          AZ_APP_NAME@stage-secondary: ''
           AZ_APP_NAME@demo: ''
           AZ_APP_NAME@uat: ''
           AZ_APP_NAME@prod: ''
+          AZ_APP_NAME@prod-secondary: ''
+          # TODO: Uncomment if you don't use an MSI with your app service
```
mike-schenk, line 139: That comment seems backward.... shouldn't it be "Uncomment if you _do_ use an MSI?"

And in fact, I can't think of any case where a repository that's prepared for this kind of deployment to a secondary region for failover _wouldn't_ use user-assigned MSIs, so probably everything about MSIs should _not_ be commented-out in this template.

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -103,27 +110,71 @@ jobs:
...
           APP_INSIGHTS_NAME@dev: ''
           APP_INSIGHTS_NAME@qa: ''
-          APP_INSIGHTS_NAME@stage: ''
+          APP_INSIGHTS_NAME@stage stage-secondary: '' # Currnetly appinsights isn't available in West Central US and you have to use West US 2 AppInsights
           APP_INSIGHTS_NAME@demo: ''
           APP_INSIGHTS_NAME@uat: ''
-          APP_INSIGHTS_NAME@prod: ''
+          APP_INSIGHTS_NAME@prod prod-secondary: '' # Currnetly appinsights isn't available in West Central US and you have to use West US 2 AppInsights
+          # Variables used for referencing secrets
+          PRIMARY_ENVIRONMENT@dev qa stage prod demo: '${{ github.event.inputs.environment }}'
```
mike-schenk, line 155: This variable name confused me at first. Maybe it would be better if it were something like `ENVIRONMENT_FOR_GITHUB_SECRETS`

mike-schenk (commented):
I'm starting to think the workflow input name `environment` and the env variable `ENVIRONMENT` that holds its value could be better named now that we're introducing the secondaries.

"stage-secondary" and "prod-secondary" aren't actually environments... not according to GitHub environments, and not according to our other uses of the term "environment". I think "destination" or "deploy-target" or something similar would be better and help avoid some of the confusion that might arise around the GitHub environment (where secrets are held) and the "environment" that's used to determine configuration values (like domain names and dependency locations).

jsclifford:
> I'm starting to think the workflow input name `environment` and the env variable `ENVIRONMENT` that holds its value could be better named now that we're introducing the secondaries.
> 
> "stage-secondary" and "prod-secondary" aren't actually environments... not according to GitHub environments, and not according to our other uses of the term "environment". I think "destination" or "deploy-target" or something similar would be better and help avoid some of the confusion that might arise around the GitHub environment (where secrets are held) and the "environment" that's used to determine configuration values (like domain names and dependency locations).

The start-project uses primary and secondary environments in the terraform.  Using this naming convention, I wrote this workflow and tested it.

workflow-templates/im-deploy-az-app-manually.yml
jsclifford: Updated to GITHUB_SECRETS_ENVIRONMENT

workflow-templates/im-deploy-az-app-manually.yml
jsclifford: Fixed it.  Yes I originally had delete if you don't use it.

workflow-templates/im-deploy-az-app-manually.yml
jsclifford: Updated to TARGET_RESOURCE_GROUP and comment.

jsclifford:
Related pull request with getting-started guide.  
https://github.com/im-practices/getting-started/pull/100

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -136,7 +136,7 @@ jobs:
           AZ_APP_NAME@uat: ''
           AZ_APP_NAME@prod: ''
           AZ_APP_NAME@prod-secondary: ''
-          # TODO: Uncomment if you don't use an MSI with your app service
+          # TODO: Uncomment if you use an (User Managed Identity) MSI with your app service
```
mike-schenk, line 139:
```suggestion
          # TODO: Uncomment if you use a (User Assigned) MSI with your app service
```

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -103,27 +110,71 @@ jobs:
...
+          # AZ_APP_MSI@qa: ''
+          # AZ_APP_MSI@stage stage-secondary: ''
+          # AZ_APP_MSI@demo: ''
+          # AZ_APP_MSI@uat: ''
+          # AZ_APP_MSI@prod prod-secondary: ''
           # This variable is used to annotate app insights
           APP_INSIGHTS_NAME@dev: ''
           APP_INSIGHTS_NAME@qa: ''
-          APP_INSIGHTS_NAME@stage: ''
+          APP_INSIGHTS_NAME@stage stage-secondary: '' # Currnetly appinsights isn't available in West Central US and you have to use West US 2 AppInsights
```
mike-schenk, line 150:
```suggestion
          APP_INSIGHTS_NAME@stage stage-secondary: '' # Currently appinsights isn't available in West Central US and you have to use West US 2 AppInsights
```

mike-schenk:
> > I'm starting to think the workflow input name `environment` and the env variable `ENVIRONMENT` that holds its value could be better named now that we're introducing the secondaries.
> > "stage-secondary" and "prod-secondary" aren't actually environments... not according to GitHub environments, and not according to our other uses of the term "environment". I think "destination" or "deploy-target" or something similar would be better and help avoid some of the confusion that might arise around the GitHub environment (where secrets are held) and the "environment" that's used to determine configuration values (like domain names and dependency locations).
> 
> The start-project uses primary and secondary environments in the terraform. Using this naming convention, I wrote this workflow and tested it. If I diverge from the start-project, then that repo and all documentation needs to be updated as well to be consistent.
> 
> I think diverging from environment will need a lot more training that using stage-secondary and prod-secondary environment. Although it is not technically correct. It is the most intuitive and easiest to adopt for teams. Not having a second input is simpler than checking a box for secondary deploy. Or instead of using primary and secondary, we use primary and failover. Then it is clear. But teams that use the secondary environment as a hot site, it doesn't apply. I though secondary is a better term that encompasses both scenarios.

Sorry, I might have been a little unclear.

I like the terms `stage-secondary` and `prod-secondary`. I'm just suggesting that the workflow input currently named `environment` that holds one of those values would be better named `target`, `deploy-target`, `destination` or similar, and the environment variable named `ENVIRONMENT` would be better named `TARGET`, `DEPLOY_TARGET`, `DESTINATION` or similar. This naming should align with the name of the environment variable that holds the name of the (target|destination) Resource Group.

Another thing I think would make sense—specifically for the Terraform workflows—would be to use an input named `root-module` instead of `environment`. The options values would still be `dev`, ... , `stage`, `prod`, `stage-secondary`, `prod-secondary`. This would correspond with the [terminology of Terraform](https://developer.hashicorp.com/terraform/language/modules#the-root-module) and its values would match the names of the respective directories in the repository.

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -44,6 +44,8 @@ on:
           - demo
           - uat
           - prod
+          - stage-secondary
+          - prod-secondary
```
bradyclifford, line 48: I recall asking @danielle-casella-adams  this question a few years ago on whether or not the secondary failover regions are represented as environments.  The response I received is they should not, rather as a separate identifier.

Has that now changed?

bradyclifford: With that said, I don't see it a problem including that second identifier as part of the environment.

Jazun713: I'm not sure how you could deploy to a secondary region without a pseudo environment of some kind.  Unless everything gets deployed at the same time (web-app, networking, etc) I think you need a logical way to separate that.

Jazun713 (commented):
seems fine to me.

workflow-templates/im-deploy-az-app-manually.yml
mike-schenk: I agree with Danielle-of-a-few-years-ago 😉 that the secondary regions are not "environments". But, as [I commented earlier](https://github.com/im-practices/.github/pull/194#issuecomment-1400760897), we can still have a consolidated list of options with `dev`, `stage`, etc. plus `stage-secondary` and `prod-secondary` like Josh has done here if we just rename that input to something other than "environment".

mike-schenk: > I'm not sure how you could deploy to a secondary region without a pseudo environment of some kind. Unless everything gets deployed at the same time (web-app, networking, etc) I think you need a logical way to separate that.

Agreed. I'm suggesting we call it "target" or "destination" or something like that in these deployment workflows.

danielle-casella-adams: Like Mike said they aren't separate environments they are a subset of the resources in an environment.  If you are deploying though you do need a way to distinguish between your regular stage resources and your stage secondary like Jason pointed out.  It makes it a little weird if we change the input name though because we are setting the GitHub Actions environment based on that input.

If I was setting up a secondary region for my azure resources I wouldn't want to create another GitHub Actions environment that has duplicate values (ARM_CLIENT_ID, ARM_CLIENT_SECRET, etc) and that the SP credential rotation workflow doesn't know how to sync to.  It seems like an additional input is warranted, one for the environment where it pulls secrets from and one for the target or destination that indicates whether you want your primary or secondary resources to go.

workflow-templates/im-run-tf-destroy.yml
```diff
@@ -11,11 +11,15 @@
 # TODO Prerequisites
 #    - Ensure each of the repo-level and env-level secrets used in this workflow have been populated by an admin in your repository.
 
-name: Terraform Destroy
+name: Terraform Destroy 💥
```
danielle-casella-adams, line 14: Not everyone uses the icons, so I probably would not include that here.
```suggestion
name: Terraform Destroy
```

workflow-templates/im-deploy-az-app-manually.yml
danielle-casella-adams: I just made my way through the rest of this file and see that it is setting a GITHUB_SECRETS_ENV based on the input.  I think in that case Mike makes a good point and changing the name of the input is the way to go, something like target or destination.

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -71,9 +73,14 @@ jobs:
 
     outputs:
       # To use these values: ${{ needs.set-vars.outputs.<OUTPUT_NAME> }}
-      RESOURCE_GROUP: ${{ steps.set-variables.outputs.RESOURCE_GROUP }}
+      TARGET_RESOURCE_GROUP: ${{ steps.set-variables.outputs.TARGET_RESOURCE_GROUP }}
+      PRIMARY_RESOURCE_GROUP: ${{ steps.set-variables.outputs.PRIMARY_RESOURCE_GROUP }}
       AZ_APP_NAME: ${{ steps.set-variables.outputs.AZ_APP_NAME }}
+      # AZ_APP_MSI: ${{ steps.set-variables.outputs.AZ_APP_NAME }} # TODO: Uncomment if you don't use an MSI with your app service
```
danielle-casella-adams, line 79:
```suggestion
      # AZ_APP_MSI: ${{ steps.set-variables.outputs.AZ_APP_MSI }} # TODO: Uncomment if you don't use an MSI with your app service
```

workflow-templates/im-deploy-az-app-manually.yml
jsclifford: I do argue that not providing a solution is way worse than just providing a way for teams to easily adopted.  Adding another variable will make it harder to track.  I like the idea of renaming `environment` field to a combo.  `environment-and-target`.  Then it is correct in both senses, but technically correct.

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -184,7 +235,7 @@ jobs:
   deploy-code:
     needs: [set-vars, validate-tag-is-in-main-for-prod-deploys]
     runs-on: [self-hosted, ubuntu-20.04]
-    environment: ${{ github.event.inputs.environment }} # Use inputs context because env context is not available to environment:
+    environment: ${{ needs.set-vars.outputs.GITHUB_SECRETS_ENVIRONMENT }} # Use inputs context because env context is not available to environment:
```
danielle-casella-adams, line 238:
```suggestion
    environment: ${{ needs.set-vars.outputs.GITHUB_SECRETS_ENVIRONMENT }} 
```

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -103,27 +110,71 @@ jobs:
...
+          TARGET_RESOURCE_GROUP@uat: ''
+          TARGET_RESOURCE_GROUP@prod: ''
+          TARGET_RESOURCE_GROUP@prod-secondary: ''
+
+          # Resource group holding the state storage account and managed service identities
+          PRIMARY_RESOURCE_GROUP@dev: ''
+          PRIMARY_RESOURCE_GROUP@qa: ''
+          PRIMARY_RESOURCE_GROUP@stage stage-secondary: ''
+          PRIMARY_RESOURCE_GROUP@demo: ''
+          PRIMARY_RESOURCE_GROUP@prod prod-secondary: ''
```
danielle-casella-adams, line 128: Including these items seems like kind of an advanced scenario that introduces a lot of extra work in this workflow.  Do we think this is something a lot of others use or is it specific to your team?

At least as far as the comments go for both blocks, I think most teams have the state in their "target_resource_group".

workflow-templates/im-deploy-tf-manual-apply.yml
```diff
@@ -147,7 +172,7 @@ jobs:
   tf-plan:
     needs: [set-vars, validate-tag-is-in-main-for-prod-deploys]
     runs-on: [self-hosted, ubuntu-20.04]
-    environment: ${{ github.event.inputs.environment }} # Use inputs context because env context is not available to environment:
+    environment: ${{ needs.set-vars.outputs.GITHUB_SECRETS_ENVIRONMENT }} # Use inputs context because env context is not available to environment:
```
danielle-casella-adams, line 175:
```suggestion
    environment: ${{ needs.set-vars.outputs.GITHUB_SECRETS_ENVIRONMENT }}
```

workflow-templates/im-deploy-tf-manual-apply.yml
```diff
@@ -314,6 +371,25 @@ jobs:
       - name: Terraform Apply
         run: terraform apply -auto-approve -no-color -lock-timeout=90s -input=false ./plans/tfplan
 
+      # TODO: Uncommnet to Annotate in AppInsights
```
danielle-casella-adams, line 374:
```suggestion
      # TODO: Uncomment to Annotate in AppInsights
```

workflow-templates/im-run-tf-destroy.yml
```diff
@@ -27,14 +31,17 @@ on:
...
+          - stage-secondary
+          - prod-secondary
+      tf-targets:
+        description: 'Targeted Destroys: format -target module.your-module.resource'
+        required: false
 
 env:
   ENVIRONMENT: ${{ github.event.inputs.environment }}
-  GITHUB_REF: ${{ github.event.inputs.branch-tag-sha  }}
+  GITHUB_REF: ${{ github.event.inputs.branch-tag-sha  }} # This is the branch/tag/sha that we'll be deploying
```
danielle-casella-adams, line 42:
```suggestion
  GITHUB_REF: ${{ github.event.inputs.branch-tag-sha  }} # This is the branch/tag/sha where terraform resources will be destroyed
```

workflow-templates/im-run-tf-destroy.yml
```diff
@@ -11,11 +11,15 @@
 # TODO Prerequisites
```
danielle-casella-adams, line 11: This workflow is meant to be a very lightweight runbook for destroying a resource, why do we need to expand it to do the stakeholder/attestor approvals and do a terraform apply and update a deployment board?

workflow-templates/im-deploy-az-app-manually.yml
bradyclifford: Another option is to continue to use the single input, but break out the two meanings into ENV vars.  I have been doing this with my workflows using a simple action I created.

This example is more complex and serving other purposes, but it shows how it can be used:
1. The target input represents more than just the environment: https://github.com/im-enrollment/ifp-enroll-connector-infrastructure/blob/main/.github/workflows/im-deploy-tf-manual-apply.yml#L16

2. GitHub action that parses out the additional values represented in the single input: https://github.com/im-enrollment/ifp-enroll-connector-infrastructure/blob/main/.github/workflows/im-deploy-tf-manual-apply.yml#L116

3. The rest of the workflow is able to use the environment variable separate from the location variable:
https://github.com/im-enrollment/ifp-enroll-connector-infrastructure/blob/main/.github/workflows/im-deploy-tf-manual-apply.yml#L66

That way I am able to still able to target the GitHub Environment while loading up my resource configuration for the secondary target.

The GitHub Action I am using to parse out those values:
https://github.com/im-enrollment/parse-scopes-action

workflow-templates/im-deploy-az-app-manually.yml
mike-schenk: When the development/production pipeline creates resource groups, it creates only a single storage account (for holding tfstate) in the primary region's resource group and it makes that storage account GRS (or GZRS), so that we can explicitly fail it over to the paired region, making it available again at its original URL, but now in the paired region.

So, actually most teams have their state in their primary resource group.

danielle-casella-adams: See I'm confused about what the difference is for each of these resource groups.  Maybe some different naming would help clear things up?

jsclifford: So it relates to the primary and secondary methodology for stage and prod.  There is a primary environment NA26 and a failover environment na27.  The target is the one you are deploying to.  The Primary is the one that has the geo redundant resources such as the storage account.  I originally had `RESOURCE_GROUP` and `PRIMARY_RESOURCE_GROUP` in this pull request.   I think the comments above explain what they are for.

workflow-templates/im-run-tf-destroy.yml
jsclifford: I added this so you can approve what you are destroying.  My team suggested stake holder approval, but I can remove that and just have Terraform approval.  The deployment board gives status of the destroy.  The bare bones one I used and was lacking with approvals.  It just destroys stuff.  No step to validate is very dangerous if you select the wrong environment.

jsclifford: @danielle-casella-adams  would you prefer I just create a second workflow call `im-run-tf-destroy-approvals.yml` so we have both.

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -33,8 +33,8 @@ on:
       tag:
         description: The tag for the release that will be deployed.  For Production, only tags reachable by the default branch will be accepted.
         required: true
-      environment:
-        description: The environment to deploy to
+      environment-or-target:
```
mike-schenk, line 36: I can tolerate this name.

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -181,7 +181,7 @@ jobs:
   stakeholder-approval:
     needs: [set-vars]
     runs-on: ubuntu-latest # Force this to run on github-hosted runner by using a tag that does not exist on self-hosted runners
-    environment: '${{ github.event.inputs.environment }} Stakeholder Approval' # Use inputs context because env context is not available to environment:
+    environment: '${{ github.event.inputs.environment-or-target }} Stakeholder Approval' # Use inputs context because env context is not available to environment:
```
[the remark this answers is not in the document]
jsclifford: Oh right.  Let me see if it was intentional.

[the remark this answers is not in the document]
jsclifford: It was intentional, but I think you are right.  It should have the same approvals as stage and prod.  If in a region failover, then this approval would need to be there.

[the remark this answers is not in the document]
jsclifford: Ok updated on all workflows that have stakeholder approval.

workflow-templates/im-deploy-az-app-manually.yml
jsclifford: I have added little more description to the `TARGET_RESOURCE_GROUP` comment @danielle-casella-adams

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -103,34 +110,78 @@ jobs:
...
           create-output-variables: true
         env:
-          # This variable is used to swap slots and annotate app insights
-          RESOURCE_GROUP@dev: ''
-          RESOURCE_GROUP@qa: ''
-          RESOURCE_GROUP@stage: ''
-          RESOURCE_GROUP@demo: ''
-          RESOURCE_GROUP@uat: ''
-          RESOURCE_GROUP@prod: ''
+          # Resource group you are targeting for deploy.  Also this variable is used to delete and re-create azure locks.
```
danielle-casella-adams, line 113: What if these comments above the target and primary indicated they should include primary and secondary regions for target which are generally na26 and na27, and the second comment indicated it should generally be the na26 region.  I know if you sit and look at the values you can fill out you would figure that out, but I did that yesterday and still got them backwards.

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -71,9 +73,14 @@ jobs:
 
     outputs:
       # To use these values: ${{ needs.set-vars.outputs.<OUTPUT_NAME> }}
-      RESOURCE_GROUP: ${{ steps.set-variables.outputs.RESOURCE_GROUP }}
+      TARGET_RESOURCE_GROUP: ${{ steps.set-variables.outputs.TARGET_RESOURCE_GROUP }}
+      PRIMARY_RESOURCE_GROUP: ${{ steps.set-variables.outputs.PRIMARY_RESOURCE_GROUP }}
       AZ_APP_NAME: ${{ steps.set-variables.outputs.AZ_APP_NAME }}
+      # AZ_APP_MSI: ${{ steps.set-variables.outputs.AZ_APP_MSI }} # TODO: Uncomment if you don't use an MSI with your app service
```
danielle-casella-adams, line 79:
```suggestion
      # AZ_APP_MSI: ${{ steps.set-variables.outputs.AZ_APP_MSI }} # TODO: Uncomment if you use an MSI with your app service
```

workflow-templates/im-deploy-az-app-manually.yml
```diff
@@ -141,7 +192,7 @@ jobs:
   attestor-approval:
     needs: [set-vars]
     runs-on: ubuntu-latest # Force this to run on github-hosted runner by using a tag that does not exist on self-hosted runners
-    environment: '${{ github.event.inputs.environment }} Attestor Approval' # Use inputs context because env context is not available to environment:
+    environment: '${{ github.event.inputs.environment-or-target }} Attestor Approval' # Use inputs context because env context is not available to environment:
```
danielle-casella-adams, line 195: I think this should match the environment the workflow is using above.  For the approval groups they should be using the input or the github secrets env, not a mix.

workflow-templates/im-deploy-tf-manual-apply.yml
```diff
@@ -61,8 +63,12 @@ jobs:
     runs-on: ubuntu-latest # Force this to run on github-hosted runner by using a tag that does not exist on self-hosted runners
 
     outputs:
-      RESOURCE_GROUP: ${{ steps.set-variables.outputs.RESOURCE_GROUP }} # To use this value: ${{ needs.set-vars.outputs.RESOURCE_GROUP }}
-      STORAGE_ACCOUNT: ${{ steps.set-variables.outputs.STORAGE_ACCOUNT }} # To use this value: ${{ needs.set-vars.outputs.STORAGE_ACCOUNT }}
+      # To use these values: ${{ needs.set-vars.outputs.<OUTPUT_NAME> }}
+      TARGET_RESOURCE_GROUP: ${{ steps.set-variables.outputs.TARGET_RESOURCE_GROUP }}
+      PRIMARY_RESOURCE_GROUP: ${{ steps.set-variables.outputs.PRIMARY_RESOURCE_GROUP }}
+      STORAGE_ACCOUNT: ${{ steps.set-variables.outputs.STORAGE_ACCOUNT }}
+      # APP_INSIGHTS_NAME: ${{ steps.set-variables.outputs.APP_INSIGHTS_NAME }} # TODO: Uncommnet to Annotate in AppInsights
```
danielle-casella-adams, line 70:
```suggestion
      # APP_INSIGHTS_NAME: ${{ steps.set-variables.outputs.APP_INSIGHTS_NAME }} # TODO: Uncomment to Annotate in AppInsights
```

danielle-casella-adams (approved):

danielle-casella-adams (changes requested):
There are a few comments in my review that need to be updated.

workflow-templates/im-deploy-az-app-manually.yml
danielle-casella-adams: I'm not sure how I feel about the environment-or-target name.  The description says the environment or target to deploy to.  That kind of makes it seem like it's either or, but it's just a subset of the environment.  I haven't looked at it myself but would providing a second input that targets the region (primary/secondary) make this clearer rather than modifying the environment?

mike-schenk: An additional input to select the region isn't quite right either, because only Stage and Prod environments have both regions as an option. Dev, QA, Stage, and Demo only have the primary region.

This is why I favored `destination`. That way the variable _name_ doesn't imply anything about whether it's a GitHub environment from which to get secrets, whether it's a primary or secondary or any other aspect. It only implies that it's a known destination, whether that destination is a whole single-region environment, whether its the primary region of an environment, the secondary region of an environment, or even a customized workflow might deploy to multiple regions of an environment simultaneously. It's just as reasonable, in my mind, to say "the deploy destination is 'dev'" as it is to say "the deploy destination is 'stage-secondary'".

Another possible name for this input could just be `deploy-to`.

workflow-templates/im-deploy-az-app-manually.yml
mike-schenk: It looks like these variables didn't get renamed to `TARGET_RESOURCE_GROUP` at the same time the corresponding ones in the `...tf-manual-apply` workflow did.

workflow-templates/im-deploy-az-app-manually.yml
jsclifford: Good catch. Only saw stakeholder in other tf workflows.

workflow-templates/im-deploy-az-app-manually.yml
jsclifford: I will update the comment to suggest the na26/na27 regions

workflow-templates/im-deploy-az-app-manually.yml
jsclifford: I have updated the description.  I feel if we aren't using environment, then there should not be any variables referencing environment in all the repos. Why do we use environment at all.  Not having environment will not be intuitive to all users.  I conceded to add target, but teams will call it an environment regardless.  They are used to the environment terminology.  I think there needs to be some give here instead being 100% syntactically correct with this.

jsclifford:
> There are a few comments in my review that need to be updated.

I have updated to addess your request changes @danielle-casella-adams.  Thank you for your input.

PR #201, opened 2023-03-14
ARCH-1808 - Creating a TechDocs publish workflow

danielle-casella-adams:


workflow-templates/im-deploy-techdocs-site.yml
```diff
@@ -0,0 +1,66 @@
...
+# TODO:  Update the triggers where TechDocs will be published
+on:
+  # push:
+  # workflow_dispatch:
+  # pull_request:
+  #   types: [closed]
+
+jobs:
+  publish-techdocs-site:
+    runs-on: [self-hosted, im-techdocs]
```
mike-schenk, line 24: Is this `im-techdocs` runner going to be present in all of our organizations?
If so, is it a separate copy for each? If so, how do we keep them always the same?

Should there be a link here that lets the reader go see what MkDocs plugins get pre-installed on that runner?

I think there should also be instructions in this file about how to install additional MkDocs plugins.

JosephDSchwartz (approved):

workflow-templates/im-deploy-techdocs-site.yml
danielle-casella-adams: I think we would have a copy of this runner in each org.  I can talk to infra-purple about whether we can make this a "global" runner that gets copied to every org when there are changes, otherwise we would have to keep it up to date in each org.

danielle-casella-adams: I chatted with Aric W and Eric J about having this as a "global" runner.  They indicated this work wasn't done yet but they didn't see a reason why it couldn't be a cross-org "global" type of runner.

I added some comments with a link to the im-practices version of the techdocs runner and what additional plugin installs look like.

mike-schenk (approved):

PR #202, opened 2023-04-25
Add reusable workflows

bradyclifford:
Reusable workflow that can be used as a job in various team workflows.

Note that it is a requirement that they reside within the `.github/workflows` directory.

Note one problem with placing them in `.github` repo is they will need to include tagged versions so they can be referenced by other repos.

Another example is having these reusable workflows live more closely to what they are using.
https://github.com/im-open/is-release-production-ready/pull/8

### DIFF
--- a/.github/workflows/reusable-examine-triggers.yml
+++ b/.github/workflows/reusable-examine-triggers.yml
@@ -2,15 +2,11 @@
 
 # Example Usage in a repo's workflow:
 # jobs:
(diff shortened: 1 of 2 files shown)


danielle-casella-adams:
I'm going to merge this into another branch and combine it with some other work that we're doing.  There might be a couple of tweaks to these files as we move forward.

danielle-casella-adams (approved):

PR #205, opened 2023-05-10
Remove the legacy auth protocal

bradyclifford:
Remove the unnecessary legacy auth protocol and additional request when downloading or uploading terraform plan output.

The `Storage Blob Data Owner` role has ben added to each BC resource group SP.
https://github.com/im-platform/development/pull/1000


mike-schenk (commented):
The version number in line 1 needs to be incremented.

Otherwise, this looks o.k. to me.

danielle-casella-adams (approved):

danielle-casella-adams:
> The version number in line 1 needs to be incremented.
> 
> Otherwise, this looks o.k. to me.

I went ahead and updated the version.  I also updated the terraform destroy workflow to use `auth-mode login` since it also used the key like the first workflow.

danielle-casella-adams (approved):

PR #208, opened 2023-05-11
SRE-250 - K6 Operator - Updating login of kubectl

jsclifford:
* Updated login method of kubectl to use azure cli.


hpractv (approved):
👍🏽

PR #209, opened 2023-05-12
ARCH-1831 - Define and implement the setup-deployment-workflow reusable job

danielle-casella-adams:


.github/workflows/im-reusable-setup-deployment-workflow.yml
```diff
@@ -0,0 +1,100 @@
...
+        required: true
+      deployment-environment:
+        description: 'The environment that is being deployed to: dev, qa, stage, stage-secondary, uat, demo, prod, prod-secondary.'
+        required: true
+        type: string
+      production-environments:
+        description: Comma separated list of production environments to check against.  Defaults to 'prod,prod-secondary'
+        required: false
+        type: string
+        default: 'prod,prod-secondary'
```
bradyclifford, line 32: Should `demo` be considered a production env. as well?

.github/workflows/im-reusable-setup-deployment-workflow.yml
```diff
@@ -0,0 +1,100 @@
...
+
+    steps:
+      - run: |
+          echo $'
+          | Deployment Arguments   |  Value                                              |
+          | ---                    |  ---                                                |
+          | Deployment Environment | `${{ inputs.deployment-environment }}`              |
+          | Ref to Deploy          | `${{ inputs.ref-to-deploy }}`                       |
+          | Actor                  | `${{ github.actor }}`                               |
+          | Workflow Source        | `${{ github.ref_name }}` - SHA: `${{ github.sha }}` |' >> $GITHUB_STEP_SUMMARY'
```
bradyclifford, line 56: Can this be toggled?  I know my team and others are recording their deployment arguments in a different way or in addition to and would like to customize this.

hpractv (approved):
I do like how this makes the rest of the consuming workflows a bit simpler to read through.

.github/workflows/im-reusable-setup-deployment-workflow.yml
danielle-casella-adams: I'm not sure about that environment.  I would think if the team wants to include demo they can override the default here.

.github/workflows/im-reusable-setup-deployment-workflow.yml
danielle-casella-adams: I can update this so you can override the summary value.  I'll include that in the next PR and make sure that workflow also does it in a toggle-able way.

PR #218, opened 2023-06-08
ARCH-1872 - Secretless access & deploys

danielle-casella-adams:
- Updating workflow templates to use secretless deployments and az cli calls


hpractv (approved):
👍🏽

PR #219, opened 2023-06-09
SRE-259 - Adding kubectl and kubelogin install steps

jsclifford:
* This will not require the runner to have kubectl and kubelogin


hpractv (approved):
👍🏽

PR #222, opened 2023-06-19
Reusable workflow fixes

danielle-casella-adams:
- Fix output value for is merge to main in setup build workflow
- Fix referenced IS_PRERELEASE output in build nuget package workflow


hpractv (approved):
👍🏽

PR #224, opened 2023-07-05
ARCH-1873 - Update templates for secret-less deploys

danielle-casella-adams:
- Remove extraneous az logins/logouts (not needed unless using az cli)
- Update terraform version
- Remove ARM_USE_OIDC and rely on it being set in terraform


hpractv (approved):
👍🏽

PR #228, opened 2023-07-25
ARCH-1907 - Remove instances of set-output

danielle-casella-adams:
Updating workflows to comply with GitHub's request to deprecate the `set-output` & `save-state` command in workflows/actions.


hpractv (approved):
👍🏽

PR #233, opened 2023-08-17
Pre-zip artifacts

danielle-casella-adams:
- Pre-zip artifacts that will be uploaded
- Remove storyteller workflow since it is unused


hpractv (approved):
👍🏽

PR #235, opened 2023-09-19
ARCH-1919 - Updating cache check action

danielle-casella-adams:
- The official actions/cache action now supports a lookup-only arg.  Replace the check-for-cache with actions/cache.  
- Take the opportunity to make it cross-platform as well.
- Change the conditions.  The actions/cache action only has an output if the cache-hit is true.  If the cache-hit is false there are no outputs.  This is weird in my opinion but it means the condition needs to change.


hpractv (approved):
👍🏽

PR #237, opened 2023-10-04
Update im-run-tf-import.yml

michael-guthrie:
Moved Checkout step before bash step, to fix issue where WORKING_FOLDER does not exist because it has not yet been loaded by Checkout.

### DIFF
--- a/workflow-templates/im-run-tf-import.yml
+++ b/workflow-templates/im-run-tf-import.yml
@@ -74,6 +74,11 @@ jobs:
         working-directory: '${{ env.TF_WORKING_DIR }}'
 
     steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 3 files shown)


mike-schenk (commented):
I don't see how this change could fix any problem. I don't see anything in the run step that could depend on the checkout.

Can you show an example of such a problem?

michael-guthrie:
> I don't see how this change could fix any problem. I don't see anything in the run step that could depend on the checkout.
> 
> Can you show an example of such a problem?

https://github.com/im-funding/funding-communication-infrastructure/actions/runs/6411365055/job/17406646574#step:2:31

Essentially the `working-directory: '${{ env.TF_WORKING_DIR }}'` doesn't represent a folder that exists before the code itself is pulled down.

michael-guthrie:
These three objects look to have been deleted altogether.

PR #239, opened 2023-10-27
WE-19955 - Adding runs-on and deploy-label

jsclifford:
* Adding runs-on input - thus allowing teams to pick the runner and decreases load in the im-linux runner.
* Added deploy label for custom labels with updated deployment board action

### DIFF
--- a/.github/workflows/im-reusable-finish-build-workflow.yml
+++ b/.github/workflows/im-reusable-finish-build-workflow.yml
@@ -7,6 +7,7 @@
 #  setup-deployment-workflow:
 #    uses: im-practices/.github/.github/workflows/im-reusable-finish-build-workflow.yml@v2
 #    with:
+#      runs-on: im-linux
 #      next-version: ${{ needs.build-deployment-artifacts.outputs.NEXT_VERSION }}
 #      title-of-teams-post: 'My CI Build'
 #      is-merge-to-main: ${{ needs.examine-triggers.outputs.IS_MERGE_TO_MAIN }}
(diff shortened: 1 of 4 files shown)


jsclifford:
Successfully ran branch against shopping pipeline.  Works fine. 
* With deployable-label input.  https://github.com/im-enrollment/shopping/actions/runs/6670194522/job/18129617111
* Without deployable-label input https://github.com/im-enrollment/shopping/actions/runs/6669363688

.github/workflows/im-reusable-finish-build-workflow.yml
```diff
@@ -32,6 +33,11 @@
 on:
   workflow_call:
     inputs:
+      runs-on:
+        description: 'The runner that this workflow will run on.'
+        required: false
+        type: string
+        default: 'im-linux'
```
danielle-casella-adams, line 40: I'm curious what the rational behind this change is.  I'm not opposed to it but these are all pretty straightforward tasks so running on different runners shouldn't be a necessity.  The im-linux runners have recently had their max number bumped so we shouldn't be running out of those runners when these builds come up.

danielle-casella-adams: ...I guess what I'm asking is, do we have a different problem we need to address?

jsclifford: I have seen these wait on the im-linux runner far longer than my custom runners.  This change will not affect any current workflows.  I was going to fork your reusable workflows to change the runners, but this is far better.  I think having the option to change the runner will have better adoption. The default is still `im-linux`.  My team and other teams would like to have the option to change the runner.

danielle-casella-adams (approved):

PR #242, opened 2023-12-22
Updates for GH actions runner changes

hpractv:
GitHub actions runners don't have the ability to pre-load PowerShell modules used int Azure SQL access and role assignments.  This adds the module installation to the template workflows.

### DIFF
--- a/workflow-templates/im-deploy-tf-auto-apply-main-to-dev-on-merge.yml
+++ b/workflow-templates/im-deploy-tf-auto-apply-main-to-dev-on-merge.yml
@@ -101,21 +101,12 @@ jobs:
           tenant-id: ${{ vars.ARM_TENANT_ID }}
           subscription-id: ${{ vars.ARM_SUBSCRIPTION_ID }}
           client-id: ${{ vars.ARM_CLIENT_ID }}
-          enable-AzPSSession: true
 
       - name: Setup Terraform
         id: setup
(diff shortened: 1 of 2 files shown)


workflow-templates/im-deploy-tf-auto-apply-main-to-dev-on-merge.yml
```diff
@@ -74,7 +74,34 @@ jobs:
...
+        run: |
+          $modules = @("SqlServer", "Az", "Az.Accounts")
+          foreach ($module in $modules) {
+            if (Get-Module -ListAvailable -Name $module) {
+              Write-Host "$module Module exists"
+            }
+            else {
+              Write-Host "$module Module does not exist"
+            }
+          }
```
mike-schenk, line 96: I think only the SqlServer module needs to be installed at this point. The Az and Az.Accounts modules aren't used in the Terraform or anything else I see.

And won't the `Install PS Modules` step fail if it is unable to install? If not, we should try to make it do so and remove this `Confirm...` step. If we can't make the install step fail on failure, then at least this Confirm step should fail if it discovers something didn't get installed. As written, it will never fail.

hpractv: Sounds good to me.

mike-schenk (approved):

PR #245, opened 2024-01-23
ARCH-2007 - Update the way outputs are set in pwsh

danielle-casella-adams:
There was an issue in 1up-infrastructure when it tried to download a blob because the plan name output it was using was empty. 
 In the plan job, the plan blob name was being set in pwsh with an echo statement.  When I changed that echo statement to the pwsh syntax the plan name output was populated and the apply job was able to download the plan blob.

There wasn't an issue using the echo statements with pwsh to set outputs before but GitHub may have made a change (or maybe our self-hosted runners changed) so that it no longer works.

```sh
echo "tf_plan_name=$terraformPlanName" >> $GITHUB_OUTPUT # No longer works
"tf_plan_name=$terraformPlanName" | Out-File -FilePath $env:GITHUB_OUTPUT -Append # This syntax works
```

### DIFF
--- a/workflow-templates/im-build-db-ci.yml
+++ b/workflow-templates/im-build-db-ci.yml
@@ -1,4 +1,4 @@
-# Workflow Code: GiddyBuzzard_v17    DO NOT REMOVE
+# Workflow Code: GiddyBuzzard_v18    DO NOT REMOVE
 # Purpose:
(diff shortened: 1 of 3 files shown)


hpractv (approved):
👍🏽. I'm sad that the syntax has gotten harder to use.  But at least you found a reliable workaround.

PR #246, opened 2024-02-02
create-github-deployment

hpractv:
Replaced update-deployment-board with create-github-deployment


danielle-casella-adams (changes requested):

danielle-casella-adams:
I forgot to add the comment to the review.  In most of the workflows the last job is called `update-deployment-board-and-send-teams-notification`.  Should that be updated?  At a minimum I think we take the `board` part out.

hpractv:
> I forgot to add the comment to the review. In most of the workflows the last job is called `update-deployment-board-and-send-teams-notification`. Should that be updated? At a minimum I think we take the `board` part out.

I'll change it to `update-github-deployments-and-send-teams-notification`.

danielle-casella-adams (approved):

PR #247, opened 2024-02-02
Remove Cleanup Deployment Board

hpractv:
We'll no longer be using the update-deployment-board and cleanup-deployment-board actions and won't need the dedicated clean up deployment board workflow


danielle-casella-adams (approved):
I meant to check this file on the previous PR and completely forgot!

PR #255, opened 2024-03-05
ARCH-2011 - Updating for node20 versions & other updates

danielle-casella-adams:
- Update actions versions to use the node20 version
- Add an input for ms-teams-uri & deploy-notifications-channel for the finish reusable workflows.  These are mainly defined as variables now but we're still treating them as secrets.  Allow caller to provide it either way. +semver:minor
- Adding a step to each reusable workflow to print out the value of each input to make it easier to see what the workflow will be using
- Add update-deployment-board permissions to the deployment workflows with a TODO to remove once we are ready to switch to techhub deploys (this affected at least one team)

### DIFF
--- a/.github/workflows/im-reusable-finish-build-workflow.yml
+++ b/.github/workflows/im-reusable-finish-build-workflow.yml
@@ -85,40 +85,50 @@ jobs:
     runs-on: ${{ inputs.runs-on }}
 
     steps:
-      - name: Check and print inputs
+      - name: Check for missing inputs
         uses: actions/github-script@v7
         with:
           script: |
-            // Some teams have this as a secret and some as a var, allow both ways
-            // but check at least one is provided
+            // Some teams have these as a secret and some as a var, allow both ways but check at least one is present
             const teamsUri = '${{ inputs.ms-teams-uri || secrets.MS_TEAMS_URI }}';
             if (!teamsUri || teamsUri.trim().length === 0){
               core.setFailed('The MS_TEAMS_URI secret or the ms-teams-uri input (preferred) must be provided.');
             }
 
-            core.info('Reusable workflow inputs:');
-            const inputs = [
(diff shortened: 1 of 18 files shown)


JosephDSchwartz (approved):
Looks good to me!

P.S. A tip for anyone else that looks at this: hide whitespace changes.

.github/workflows/im-reusable-finish-build-workflow.yml
```diff
@@ -1,168 +1,217 @@
...
+        uses: actions/github-script@v7
+        with:
+          script: |
+            // Some teams have these as a secret and some as a var, allow both ways but check at least one is present
+            const teamsUri = '${{ inputs.ms-teams-uri || secrets.MS_TEAMS_URI }}';
+            if (!teamsUri || teamsUri.trim().length === 0){
+              core.setFailed('The MS_TEAMS_URI secret or the ms-teams-uri input (preferred) must be provided.');
+            }
+
+      - name: Print inputs
```
hpractv, line 98: Do the inputs of the reusable workflow get printed in the log like when actions are called?

danielle-casella-adams: I think there are two use cases for the extra step.  First, it shows you what the action is going to use for a default if one wasn't provided which is kind of nice.  In the example below the tag-prefix isn't provided and you can see how the built in Inputs shows it versus the Print Inputs step.
image


The second use case is the built in inputs doesn't print whether it got secrets or not, which is what prompted me to do this in the first place.  Obviously the new step doesn't print the value, but it does show whether it received a value or not for the secret.
image

hpractv (approved):

danielle-casella-adams:
> P.S. A tip for anyone else that looks at this: hide whitespace changes.

Very sorry about that.  I did a format on the reusable workflows when I was done editing and there were tons of changes!

PR #259, opened 2024-04-01
Updates the Terraform auto-plan workflow

mike-schenk:
- adds -input=false to the terraform commands to avoid silent stalls in actions runs


izep (approved):
I didn't know about that parameter...

workflow-templates/im-build-tf-auto-plan-and-comment-on-prs.yml
```diff
@@ -1,126 +1,126 @@
...
-             
-            __Pusher__: @${{ github.actor }}
-            __Action__: \`${{ github.event_name }}\``;
-            github.rest.issues.createComment({
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: output
-            })
+# Workflow Code: DeterminedPorcupine_v23    DO NOT REMOVE
```
JosephDSchwartz, line 1:
```suggestion
# Workflow Code: DeterminedPorcupine_v24    DO NOT REMOVE
```

JosephDSchwartz (approved):

danielle-casella-adams:
> I didn't know about that parameter...

Same.  

[-input=true](https://developer.hashicorp.com/terraform/cli/commands/init#input-true) Ask for input if necessary. If false, will error if input was required.

I love this addition @mike-schenk!

mike-schenk:
> > I didn't know about that parameter...
> 
> Same.
> 
> [-input=true](https://developer.hashicorp.com/terraform/cli/commands/init#input-true) Ask for input if necessary. If false, will error if input was required.
> 
> I love this addition @mike-schenk!

The reason I added it here is because I had a situation where a root module required a variable that wasn't provided, so Terraform was asking for the input... but in GitHub actions, that simply appears as a stalled step that eventually just fails with a timeout and with nothing in the logs for that step.

PR #260, opened 2024-04-03
WIP - ARCH-2065 - Skip update-deployment-board step

danielle-casella-adams:
Adding a way to skip the update-deployment-board step in the im-reusable-finish-deployment-workflow without creating a breaking change yet.  There are around 70 usages of this reusable workflow and about half of those rely on the default value being 1 for the `deployment-board-number` input which is why it's a little hacky with the -1.  If we prefer a different input like `skip-deployment-board-update` could be temporarily introduced and then removed with the rest of the update-deployment-board functionality.  I don't think this will be very long-lived whichever way we go.

I added a second step that spells out why the first step might be skipped just for a little bit of additional info that can clue users into the fact that skipping is now an option.  That step can be removed once we make the breaking change to remove all update-deployment-board functionality.

This was tested in 1UP:
- [Workflow that skips update-deployment-board by setting board number to -1](https://github.com/im-practices/1UP/actions/runs/8636144478/job/23679495959#step:7:2)
- [Workflow that uses update-deployment-board by setting board number to 2](https://github.com/im-practices/1UP/actions/runs/8637434840/job/23679601756#step:6:1)

Both of those deployments are choking on the create-github-deployment step (I think because of all the current github incidents) but the update-deployment-board is happy.

### DIFF
--- a/.github/workflows/im-reusable-finish-deployment-workflow.yml
+++ b/.github/workflows/im-reusable-finish-deployment-workflow.yml
@@ -13,8 +13,8 @@
 #      title-of-teams-post: 'Deploy ${{ needs.set-vars.outputs.AZ_APP_NAME }} ${{ inputs.tag }} to ${{ inputs.environment-or-target }}'
 #      post-status-in-deployment-notifications-channel: true
 #      timezone: america/denver
-#      deployment-board-number: 1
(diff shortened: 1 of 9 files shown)


mike-schenk (approved):
Love it!

PR #264, opened 2024-05-16
Update PR comment for next version to include link to the release build

ricardozv28:


.github/workflows/im-reusable-finish-build-workflow.yml
```diff
@@ -191,7 +191,7 @@ jobs:
 
             const nextVersionText = isMergeToMain ?
               `[Tag - ${nextVersion}](https://github.com/${orgAndRepo}/releases/tag/${nextVersion})` :
-              `Next Version - ${nextVersion}`;
+              `[Next Version - ${nextVersion}](https://github.com/${orgAndRepo}/releases/${nextVersion})`;
```
hpractv, line 194: I've done some research.  In the past the `im-open/git-version-lite` action made a tag and a release at the same time.  Now they're broken into two different actions.

`im-open/git-version-lite` now creates a version, the second, `im-open/create-release`, creates the release object.

Since this reusable workflow doesn't create a release, then it's not a good place to have a link to it.

hpractv: You could create summary output in `im-open/create-release` action for the link, since we know a release is created there.

hpractv (changes requested):

.github/workflows/im-reusable-finish-build-workflow.yml
mike-schenk: I don't understand why the ternary condition is the way it is. Why should it create a link to a release when `isMergeToMain` is true, but only output the version number _without_ a link when it's false?

Is it a safe assumption that `isMergeToMain == true` is an equivalent condition to "a version tag was pushed"?

On the other hand, is it a safe assumption that if `inputs.next-version` has a value, then a tag must have been pushed for it?

In my opinion, the latter assumption is more likely to be true. In that case, the whole conditional expression would need to be different, and possibly even the default value for the input should be different.

Maybe like:
```js
const nextVersionText = nextVersion === 'N/A' ?
  `Next Version - ${nextVersion}` :
  `[Tag - ${nextVersion}](https://github.com/${orgAndRepo}/releases/tag/${nextVersion})`;
```
Or if we change the default of the next-version input to a null or empty string:
```js
const nextVersionText = nextVersion ?
  'Next Version - N/A' :
  `[Tag - ${nextVersion}](https://github.com/${orgAndRepo}/releases/tag/${nextVersion})`;
```

PR #267, opened 2024-06-13
ARCH-2088 - +semver:breaking Changes for TechHub deployments

danielle-casella-adams:
- Remove deprecated inputs from the reusable finish deployment workflow
- Make techhub deployment related inputs required in the finish deployment workflow
- Update workflow templates to use latest reusable workflow versions (@v3) & modify inputs accordingly
- Update permissions in each of the "deploy" templates now that deployment-board permissions are no longer needed

### DIFF
--- a/workflow-templates/im-run-validate-deployed-terraform.yml
+++ b/workflow-templates/im-run-validate-deployed-terraform.yml
@@ -1,4 +1,4 @@
-# Workflow Code: ShinySQUIRREL_v25    DO NOT REMOVE
+# Workflow Code: ShinySQUIRREL_v26    DO NOT REMOVE
 # Purpose:
 #    Validates that the deployed terraform matches what is supposed to be deployed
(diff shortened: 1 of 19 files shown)


workflow-templates/im-run-validate-deployed-terraform.yml
```diff
@@ -67,6 +66,7 @@ jobs:
         working-directory: '${{ env.TF_WORKING_DIR }}'
 
     steps:
+      # TODO:  SWAT - Need a replacement for determining which release is the latest
```
danielle-casella-adams, line 69: Just a heads up, I left a TODO here because I think it's kind of an involved fix.  I looked at the way IFP did it and it seems like this could be simplified.  I'd like to work on a more streamlined solution before updating this file.  Maybe part of that is investigating whether this template is providing the value we want.

hpractv (approved):
👍🏽

PR #271, opened 2024-07-17
ARCH-2122 - Adding reusable workflow for validating catalog-info.yml

danielle-casella-adams:
The PR creates a new reusable workflow for validating catalog-info.yml files.  It has been incorporated into the workflow templates in two ways: as part of the CI build or as a standalone validation workflow.

To see it in action:
## pull_request
- [File changed and `only-validate-if-file-changed=true`](https://github.com/im-practices/1UP/actions/runs/9980832184/job/27583127036#step:5:42)
- [File did not change and `only-validate-if-file-changed=true`](https://github.com/im-practices/1UP/actions/runs/9980946516/job/27583481442#step:5:42)
- [File did not change but `only-validate-if-file-changed=false`](https://github.com/im-practices/1UP/actions/runs/9980956679/job/27583514148#step:5:42)

## push
- [File changed and `only-validate-if-file-changed=true`](https://github.com/im-practices/1UP/actions/runs/9980831829/job/27583126102#step:5:42)
- [File did not change and `only-validate-if-file-changed=true`](https://github.com/im-practices/1UP/actions/runs/9980946065/job/27583479851#step:5:42)
- [File did not change but `only-validate-if-file-changed=false`](https://github.com/im-practices/1UP/actions/runs/9980956182/job/27583512823#step:5:42)

The pull_request and push runs will be the same except for the step where it tries to create a PR comment.  That will obviously never run if it's not a pull_request trigger.

### DIFF
--- a/.github/workflows/im-reusable-validate-catalog-info.yml
+++ b/.github/workflows/im-reusable-validate-catalog-info.yml
@@ -1,0 +1,159 @@
+# The purpose of this reusable job is to validate the catalog-info.yml file
+# by optionally checking if the file changed in the PR, checking out the repo,
+# validating the catalog-info.yml file using im-open/validate-catalog-info-file
+# and optionally updating the PR comment with the validation results.
+
+# Example Usage in a repo's workflow:
+# jobs:
+#  validate-catalog-info:
+#    uses: im-practices/.github/.github/workflows/im-reusable-validate-catalog-info.yml@v3
+#    with:
+#      runs-on: im-linux
+#      filename: catalog-info.yml
+#      add-pr-comment: true
+#      only-validate-if-file-changed: true
+#      fail-validation-step-if-errors: true
+#      have-validation-step-generate-job-summary: true
+
+on:
+  workflow_call:
+    inputs:
+      runs-on:
+        description: 'The runner that this workflow will run on.'
+        type: string
+        required: false
+        default: 'im-linux'
+      filename:
+        description: |
+          The name of the catalog info file that contains TechHub entities.
+          For most cases this will be catalog-info.yml.
+        type: string
+        required: false
+        default: 'catalog-info.yml'
+      add-pr-comment:
+        description: |
+          THIS ONLY APPLIES IF THE TRIGGER IS PULL_REQUEST.
+          Flag indicating whether a PR comment should be made with the validation results.
+          By default this is true so a PR comment will be made each time the workflow runs.
+        type: boolean
+        required: false
+        default: true
+      only-validate-if-file-changed:
+        description: |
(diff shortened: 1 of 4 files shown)


.github/workflows/im-reusable-validate-catalog-info.yml
```diff
@@ -0,0 +1,159 @@
...
+      only-validate-if-file-changed:
+        description: |
+          THIS ONLY APPLIES IF THE TRIGGER IS PULL_REQUEST OR PULL.
+          Flag indicating whether the validation should happen on every run or just if the catalog-info.yml file changed.
+          This is set to false by default, so validation will happen on every run.
+          If the trigger is workflow_dispatch or schedule, false is an appropriate choice because validation should always run.
+          For a trigger like pull_request you may want to set this to true so validation only happens if the file changed.
+        type: boolean
+        required: false
+        default: false
```
danielle-casella-adams, line 50: I wanted to get other's thoughts on this flag (especially @mike-schenk).  Initially Mike indicated he wanted a smart reusable workflow that only ran if it needed to.

In this job though there are a lot of steps that have to run to determine if the validation steps in this job should run.  It currently has this optional flag that makes it so the validation steps only run if the catalog-info.yml file changed, but it has to do a number of steps to examine all the applicable conditions.  Since it's already doing that work, does it make sense to bail on the last two validation/reporting steps if the file didn't change?  Does this flag just introduce complexity?  Is it worth doing?  Would turning off the failures/job summary be enough in that case?

I made a new standalone template and updated im-build-dotnet-ci with another job so they both use this reusable workflow for validating catalog-info.yml.  Inside of the ci job it runs at the same time as the other jobs, so it definitely won't be the longest item.  I'm not sure if the selective validation is worth it since we're doing work in the reusable job to figure out if it should run.  In the standalone template, the sole point is to validate the catalog-info.yml on a scheduled basis, so it shouldn't ever skip those validation steps.

mike-schenk: Nah, I'd do the file change detection outside the reusable workflow then only invoke it conditionally.

I want to propose and promote a more efficient and comprehensive way to implement required status checks on PRs— using a singular PR-triggered workflow that just figures out which other workflows are applicable to call-out to, then fails only if any of them fail. It allows PRs that don't change anything significant to pass their required status check quickly so they can be merged without waiting for a whole battery of irrelevant tests and evaluations to run.

I've implemented something I'm pretty happy with in the funding-calculation repo, which I have just now extended to include your reusable workflow: https://github.com/im-funding/funding-calculation/pull/320

mike-schenk (approved):

PR #276, opened 2024-08-26
SRE-630 - Making terraform deploys runs-on specific to a region for SQL TF deploys

jsclifford:
* Making terraform deploys runs-on specific to a region for SQL TF deploys


danielle-casella-adams (approved):

mike-schenk:
The comments could stand to be more precise. The phrase "Managed SQL Server" could be confused.

The region-specific firewall limitation applies to both "Azure SQL Database" as well as "Azure SQL Managed Instance", but does _not_ apply to "SQL Server on Azure VMs".

So, I think the term "Azure SQL" would be better than "Managed SQL Server".
image

PR #277, opened 2024-09-05
EN-23320 - Adding Teams action in reusable workflows

jsclifford:
* Adding Teams Notification custom-action as an input
* Adding version in Annotation for build reusable workflow.

### DIFF
--- a/.github/workflows/im-reusable-finish-build-workflow.yml
+++ b/.github/workflows/im-reusable-finish-build-workflow.yml
@@ -66,8 +66,17 @@ on:
         description: Additional PR comment content aside from workflow run and next version.
         required: false
         type: string
+      send-teams-notification:
+        description: Flag indicating whether to send a notification to the team's channel.  Expected `true` or `false`.
+        required: false
+        type: boolean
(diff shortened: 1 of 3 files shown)


jsclifford:
I have testing my branch on this workflow run and it is passing. https://github.com/im-enrollment/plan-loading/actions/runs/10726318332

.github/workflows/im-reusable-finish-build-workflow.yml
```diff
@@ -66,10 +66,19 @@ on:
         description: Additional PR comment content aside from workflow run and next version.
         required: false
         type: string
+      send-teams-notification:
+        description: Flag indicating whether to send a notification to the team's channel.  Expected `true` or `false`.
+        required: false
+        type: boolean
+        default: true
```
hpractv, line 73: Would having this default to true just throw an error for those teams that don't want to deploy to their teams channel, so they would have to specify `false`?

It's a negative feed back opt-in.

hpractv (approved):
👍🏽

PR #281, opened 2024-09-19
No longer require ms-teams-uri in finish deployment

danielle-casella-adams:
Make the ms-teams-uri inputs option since teams can no longer create new integrations.


hpractv (approved):
👍🏽

PR #282, opened 2024-09-19
Teams Check Fix

danielle-casella-adams:
Fix for the teams check. The `if` did not like the same syntax as the input or the js code above because `if` does not have access to the `secrets` context.
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#context-availability


hpractv (approved):
👍🏽

PR #283, opened 2024-10-14
SRE-208 - Updating workflow to output NEXT_RELEASE_VERSION.

jsclifford:
- Added outputs `NEXT_RELEASE_VERSION` and `NEXT_RELEASE_VERSION_NO_PREFIX`.  
- Updating to latest git-version-lite action

### DIFF
--- a/.github/workflows/im-reusable-setup-build-workflow.yml
+++ b/.github/workflows/im-reusable-setup-build-workflow.yml
@@ -64,14 +64,13 @@ on:
       NEXT_VERSION_NO_PREFIX:
         description: The next version calculated by git-version-lite without a prefix.
         value: ${{ jobs.setup-build-workflow.outputs.NEXT_VERSION_NO_PREFIX }}
-      NEXT_VERSION_NO_PRERELEASE:
-        description: The next version calculated by git-version-lite without a postfix.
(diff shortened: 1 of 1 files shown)


mike-schenk (changes requested):
I don't think we should be committing package.json version numbers to the repo since we determine version numbers during our builds. My reasons are basically the same as what's [documented by semantic-release here](https://semantic-release.gitbook.io/semantic-release/support/faq#why-is-the-package.jsons-version-not-updated-in-my-repository).

`start-project` is an example where the package version is not committed to source control: https://github.com/im-practices/start-project/blob/main/src/start-project/package.json

If there's another reason we need these outputs, I'll listen.

But regardless, I don't think it's worth running git-version-lite _twice_ every time this reusable workflow is used when a very small subset of usages need the extra output.

As an alternative, we could expand git-version-lite itself to output the additional values directly.

jsclifford:
Ok I will update the action to output both versions in git-version lite.

jsclifford:
I updated git-version-lite to output the bare versions.  https://github.com/im-open/git-version-lite/pull/49 please review.  Once merged I will update this workflow.

jsclifford:
@mike-schenk  please review pull request.  I have updated code per your request in your comments

mike-schenk (approved):

PR #284, opened 2024-10-28
SRE-208 - Creating playwright workflow templates

jsclifford:
* Creating playwright workflow templates

### DIFF
--- a/workflow-templates/im-test-playwright-manual.yml
+++ b/workflow-templates/im-test-playwright-manual.yml
@@ -42,7 +42,7 @@ on:
       browser-project:
         description: Browser - what browser project to use. # Browser projects defined in the playwright.config.ts file.
         type: choice
-        options:
+        options: # TODO: Update this list to match the browser projects in the playwright.config.ts file.
           - chromium
           - firefox
           - webkit
@@ -90,7 +90,7 @@ jobs:
       - name: Calculate Ref to Checkout
         id: calculate-ref
         run: |
-          GITHUB_REF="${{ github.event_name == 'pull_request' && github.head_ref || env.DISPATCH_REF }}"
+          GITHUB_REF="${{ env.DISPATCH_REF }}"
           echo "GITHUB_REF=$GITHUB_REF" >> $GITHUB_ENV
           echo "GITHUB_REF=$GITHUB_REF" >> $GITHUB_OUTPUT
 
@@ -104,8 +104,7 @@ jobs:
           | Workflow Branch/Tag     | `${{ github.ref_name }}` - SHA: `${{ github.sha }}`                                       |
           | Browser                 | `${{ inputs.browser-project }}`                                                           |
           | Playwright Config File  | `${{ inputs.config-file }}`                                                               |
-          | Test Filter             | `${{ inputs.test-filter != 0 && inputs.test-filter || 'NA' }}`                            |
-          | Send Teams Notification | `${{ inputs.send-teams-notification == 0 && 'false' || inputs.send-teams-notification }}` |' >> $GITHUB_STEP_SUMMARY
+          | Test Filter             | `${{ inputs.test-filter != 0 && inputs.test-filter || 'NA' }}`                            |' >> $GITHUB_STEP_SUMMARY
 
       - name: Checkout Repository
         uses: actions/checkout@v4
@@ -113,7 +112,7 @@ jobs:
           ref: ${{ env.GITHUB_REF }}
 
       - name: Set Cache Keys
-        # As long as these files don't change the cache will remain intact and the time
+        # As long as these files don't change the cache will remain in tact and the time
         # to restore the different packages in the various jobs will be reduced.  If one
         # of the files does change though, a new cache will be uploaded the next time
         # it runs, then subsequent runs should be faster.
@@ -128,6 +127,10 @@ jobs:
           path: |
             ${{ env.NPM_CACHE_PATH }}
 
+      # The remaining steps will only be executed if the cache was not found, otherwise they will be skipped.
+      # This action creates a post-job step that will upload the node_modules dir to the cache if the job
+      # completes successfully.  Subsequent jobs and workflow runs can use this cached version of the node_modules
+      # folder if the package-lock.json hasn't changed and it uses a ubuntu-20.04 runner to restore the cache from.
       - name: Setup Node.js
         uses: actions/setup-node@v4
         if: steps.has-cache.outputs.cache-hit != 'true'
@@ -139,7 +142,8 @@ jobs:
         uses: im-open/authenticate-with-gh-package-registries@v1
         with:
           read-pkg-token: ${{ secrets.READ_PKG_TOKEN }} # This is an org-level secret
-          orgs: 'im-client,im-enrollment,im-customer-engagement,im-practices,im-open'
+          orgs: 'im-client,im-enrollment,im-customer-engagement,im-practices,im-open' # TODO: Verify list of orgs packages will be pulled from  
+
 
       - name: NPM Install
         if: steps.has-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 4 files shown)


workflow-templates/im-test-playwright-manual.yml
```diff
@@ -0,0 +1,224 @@
...
+          - stage
+          # - prod # Uncomment this line if you want to test against prod
+      browser-project:
+        description: Browser - what browser project to use. # Browser projects defined in the playwright.config.ts file.
+        type: choice
+        options:
+          - chromium
+          - firefox
+          - webkit
+          - all
```
JosephDSchwartz, line 49: If the browser projects need to be defined in the playwright.config, does that mean the list of options here could change team by team? If so, we should add a TODO in here along the lines of `TODO: Add or remove any options that are or aren't in the playwright.config.ts for the playwright project`.

workflow-templates/im-test-playwright-manual.yml
```diff
@@ -0,0 +1,224 @@
...
+        default: '' # TODO: Add a default test filter if needed.
+
+permissions:
+  id-token: write
+  contents: read
+  actions: read
+
+env:
+  DISPATCH_REF: ${{ inputs.tag == 0 && github.ref_name || inputs.branch-tag-sha }}
+  PLAYWRIGHT_TEST_DIR: 'tests/playwright' # TODO: Update this to the path of the playwright test root door
```
JosephDSchwartz, line 70:
```suggestion
  PLAYWRIGHT_TEST_DIR: 'tests/playwright' # TODO: Update this to the path of the playwright test root directory
```

workflow-templates/im-test-playwright-manual.yml
```diff
@@ -0,0 +1,224 @@
...
+  PLAYWRIGHT_REPORT_DIR: 'tests/playwright/playwright-report' # TODO: Update this to the path of the playwright report directory
+  PLAYWRIGHT_RESULTS_DIR: 'tests/playwright/test-results' # TODO: Update this to the path of the playwright test results directory
+  TIMEZONE: America/Denver
+  NPM_CACHE_PATH: '**/tests/playwright/node_modules' # TODO: Update this to the path of the node_modules directory in the playwright test root directory
```
JosephDSchwartz, line 74: Is there a reason this start of this path is different from the other path variables above?
```suggestion
  NPM_CACHE_PATH: 'tests/playwright/node_modules' # TODO: Update this to the path of the node_modules directory in the playwright test root directory
```

workflow-templates/im-test-playwright-manual.yml
```diff
@@ -0,0 +1,224 @@
...
+
+    outputs:
+      NPM_CACHE_KEY: ${{ env.NPM_CACHE_KEY }}
+      GITHUB_REF: ${{ steps.calculate-ref.outputs.GITHUB_REF }}
+
+    steps:
+      - name: Calculate Ref to Checkout
+        id: calculate-ref
+        run: |
+          GITHUB_REF="${{ github.event_name == 'pull_request' && github.head_ref || env.DISPATCH_REF }}"
```
JosephDSchwartz, line 93: The only trigger for this template is a workflow dispatch, so the `event_name` should never be `pull_request`. Is this here just in case a team wants to include a pull_request trigger?

workflow-templates/im-test-playwright-manual.yml
```diff
@@ -0,0 +1,224 @@
...
+          | Test Filter             | `${{ inputs.test-filter != 0 && inputs.test-filter || 'NA' }}`                            |
+          | Send Teams Notification | `${{ inputs.send-teams-notification == 0 && 'false' || inputs.send-teams-notification }}` |' >> $GITHUB_STEP_SUMMARY
+
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.GITHUB_REF }}
+
+      - name: Set Cache Keys
+        # As long as these files don't change the cache will remain in tact and the time
```
JosephDSchwartz, line 116:
```suggestion
        # As long as these files don't change the cache will remain intact and the time
```

workflow-templates/im-test-playwright-manual.yml
```diff
@@ -0,0 +1,224 @@
...
+        if: steps.has-cache.outputs.cache-hit != 'true'
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Authenticate with GitHub Packages
+        if: steps.has-cache.outputs.cache-hit != 'true'
+        uses: im-open/authenticate-with-gh-package-registries@v1
+        with:
+          read-pkg-token: ${{ secrets.READ_PKG_TOKEN }} # This is an org-level secret
+          orgs: 'im-client,im-enrollment,im-customer-engagement,im-practices,im-open'
```
JosephDSchwartz, line 142: I'd probably go even as far as removing two or three of these orgs. I don't know how often teams will need all five of them.
```suggestion
          orgs: 'im-client,im-enrollment,im-customer-engagement,im-practices,im-open'  # TODO: Verify list of orgs packages will be pulled from
```

workflow-templates/im-test-playwright-manual.yml
```diff
@@ -0,0 +1,224 @@
...
+        if: steps.has-cache.outputs.cache-hit != 'true'
+        working-directory: '${{ env.PLAYWRIGHT_TEST_DIR }}'
+        run: |
+          echo "Installing NPM"
+          npm ci
+
+      # The remaining steps will only be executed if the cache was not found, otherwise they will be skipped.
+      # This action creates a post-job step that will upload the node_modules dir to the cache if the job
+      # completes successfully.  Subsequent jobs and workflow runs can use this cached version of the node_modules
+      # folder if the package-lock.json hasn't changed and it uses a ubuntu-20.04 runner to restore the cache from.
```
JosephDSchwartz, line 154: Is this comment block supposed to be over all of the steps with the `if: steps.has-cache.outputs.cache-hit != 'true'` condition?

workflow-templates/im-test-playwright-schedule.yml
JosephDSchwartz, line 1: Is this workflow basically the same as `im-test-playwright-manual` other than the workflow trigger? If so, then you could think about creating a reusable workflow (like the ones we have [here](https://github.com/im-practices/.github/tree/main/.github/workflows)) and calling that in both of these files.

JosephDSchwartz (changes requested):

workflow-templates/im-test-playwright-manual.yml
jsclifford: This ensures that it captures any node_modules.  I have this if the pathing is incorrect.  I have seen problems where the cache isn't saved when I specify the path and the working directory isn't what I expect.  I will be keeping this as it doesn't hurt to have it.  I feel it is a best practice.

workflow-templates/im-test-playwright-manual.yml
jsclifford: Added todo.

workflow-templates/im-test-playwright-manual.yml
jsclifford: I will just add the todo.  im-practices is used for the im.playwright-appinsights-reporter in an upcoming adr.

workflow-templates/im-test-playwright-manual.yml
jsclifford: Moved to above first step that has the if statement.

workflow-templates/im-test-playwright-schedule.yml
jsclifford: I can refactor this later as a reusable workflow.  I can add that to my backlog.  I am up against a timeline and don't have time to fully test the reusable workflow path.

JosephDSchwartz (approved):

PR #288, opened 2024-11-08
SRE-208 - Playwright Schedule Workflow Update

jsclifford:
- Updating workflow to use matrix that can be outputed to both the annotation and strategy job.


hpractv (approved):
👍🏽

PR #291, opened 2025-01-17
SRE-665 - Adding deployment description

jsclifford:
* Added input deployment-description to finish-workflows.

### DIFF
--- a/.github/workflows/im-reusable-finish-deployment-workflow.yml
+++ b/.github/workflows/im-reusable-finish-deployment-workflow.yml
@@ -102,6 +102,10 @@ on:
         description: The URI for the teams channel where a status will be posted.  This input is the preferred way to provide the URI but the secret should be used instead if the value is defined as a secret.
         required: false
         type: string
+      deployment-description:
+        description: 'The description of the deployment that will be used in the GitHub deployment.'
+        required: false
+        type: string
     secrets:
       MS_TEAMS_URI:
         description: 'The URI for the teams channel where a status will be posted.  The input is the preferred way to provide the URI but the secret should be used if the value is defined as a secret.'
@@ -181,7 +185,7 @@ jobs:
           environment: ${{ inputs.deployment-environment }}
           release-ref: ${{ inputs.release-tag }}
           deployment-status: ${{ steps.conclusion.outputs.workflow_conclusion }}
-          deployment-description: 'Deployment to the ${{ inputs.deployment-environment }} environment of ${{ inputs.release-tag }}'
+          deployment-description: ${{ inputs.deployment-description == 0 && format('Deployment to the {0} environment of {1}',inputs.deployment-environment,inputs.release-tag ) || inputs.deployment-description }}
           entity: ${{ inputs.entity }}
           instance: ${{ inputs.instance }}
 


jsclifford:
Testing changes to workflow in these two deployments.
* Create - https://github.com/im-enrollment/shopping-infrastructure/actions/runs/12835120528
* Destroy - https://github.com/im-enrollment/shopping-infrastructure/actions/runs/12835514536/job/35796666654

hpractv:
The status of a deployment is meant to show the disposition of the deployment not the current state of the artifacts or infrastructure.  Successfully destroying a slot or special environment is still a success. 

Instead, it would be better to add an override or refinement to the [`deployment-description`](https://github.com/im-practices/.github/blob/main/.github/workflows/im-reusable-finish-deployment-workflow.yml#L184C11-L184C48) parameter and then expose it in TechHub dashboard.

As it is, where you've labeled the instance of `NA27-destroyed`, it becomes hard to follow that environment as its own entity.
image

For a better experience and more logical status progression, we can expose the description after adding it as an updateable parameter in the reusable workflow.

jsclifford:
Once this is allowed, then I will not be adding destroyed to the instance, but just deactivating the current deployment.  Thus making the destroy as not active.  So the example I posted was not ideal.  I forgot to remove the destroyed in the instance.  

I would like a description exposed for sure.  Can I have both?

jsclifford:
> The status of a deployment is meant to show the disposition of the deployment not the current state of the artifacts or infrastructure. Successfully destroying a slot or special environment is still a success.

So I guess I am confused on the purpose of the deployment board.  I thought its purpose was to show what is active/deployed in each environment.  So the techhub deployment board becomes useless in that sense to my team.

jsclifford:
@hpractv I have updated the pull request and created a separate reusable workflow because of your concerns with adding an optional input.  Please review.

jsclifford:
I also created an issue in the github-deployments plugin for Backstage to display the deployment description.  https://github.com/im-open/im-github-deployments/issues/31

I have tested the new reusable workflow and it works as expected.  
* With deployment description - https://github.com/im-enrollment/shopping-infrastructure/actions/runs/12955011763/job/36138422696
* Without deployment description and using default description. https://github.com/im-enrollment/shopping-infrastructure/actions/runs/12954885906/job/36138169076
* Inactivate deployment set to true - https://github.com/im-enrollment/shopping-infrastructure/actions/runs/12955095155

The inactivate deployment did not uncheck the latest check box in the deployment board.  Is there another field I need to update for that?

hpractv:
We're going to make some modifications to the TechHub Plugin to add a column that shows the "Valid To" time of a deployment.  If it's never inactivated, then the column is null. If it's followed up by another deployment to the same environment/instance, then it will show its valid to time".

The description will show as a hover dialog over the Status column value.  Something like what the hover over the deployment date display does.

We're not going to add an open action that updates deployments with any (supported) status.  The deployments dashboard isn't meant to be a real time dashboard for deployments.  In the case you want to know real time results of an ongoing deployment, you really should be watching the GitHub workflow run.

It would be best to split this up into 2 PRs. One that adds the deployment description as a parameter to the first reusable workflow, which would be easy to approve, and a second for a new reusable workflow for inactivation.

The inactivation workflow is still hard to define. Especially, because the use of Teams Webhooks is likely to be deprecated making your workflow about inactivation a simple call to an im-open action.

jsclifford:
Ok thanks.  So latest is a calculated field.  We cannot directly edit that.  Well my pull request means nothing then.  I really was hoping just to update the latest field.  So thanks for all the support.  I will just update the pull request to include the description and hoping that the description will be displayed in the techhub.  Seems like your team is willing to do that. 

I updated the pull request to only add the deployment description.

hpractv:
> I would prefer that it was included as a column that cuts off after a number of characters and on hover show the full description...

I'll experiment with it. A fixed width with an ellipsis at end with a hover might work well.

> So what is the purpose of the deployment board if it doesn't show what is deployed to what environment??

It's for a state disposition environment and instances. So that we can see the history of deployments and look other BCs deployments.  If you as a specific project owner want to know what your real-time deployment status, then TechHub will never be as good or up-to-date as watching the actual deployment.

PR #298, opened 2025-04-18
Proposal: Add "Authenticate with GitHub Packages" step in other jobs for dotnet CI workflow template

jotaWTW:
While removing the Artifactory entry from **nuget.config**, I realized our .NET CI workflow template needs to include an “Authenticate with GitHub Packages” step to both the `dotnet-test` and `build-deployment-artifacts` jobs. This is required whenever a .NET app uses an internal NuGet package like the Mktp.Api.Authentication package (v2.1.14).

**Note**: I used PR [PR #251](https://github.com/im-customer-engagement/cep-case-comments/pull/251) to verify these changes.


jotaWTW:
Not required. Thanks @richard-leon-130524 the solution is use actions/cache/restore@v4 to avoid using extra Authenticate with GitHub Packages sections

PR #300, opened 2025-08-27
SRE-690 - Adding GHAS workflow template

jsclifford:
* Adding GHAS workflow template
* Includes Terraform, Javascript, Powershell, C#


jsclifford:
I added the header.

workflow-templates/im-test-code-analysis.yml
```diff
@@ -0,0 +1,432 @@
...
+          DOTNET_INSTALL_DIR: ${{ env.DOTNET_INSTALL_DIR }}
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v3
+        with:
+          ram: 3560 # Give 3.5 GB ram to CodeQL, 1024 can lead to out of memory issues for some C# repos
+          # Using additional queries to get more results
+          # See https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning#using-queries-in-ql-packs
+          queries: security-extended
```
ahernandez411, line 190: These `security-extended` lines caused a lot of extra alerts to come about. If we remove the line `queries: security-extended` it will be the base scan; which is only the higher-confidence alerts. There may be other instances of this in this workflow that I have not discovered.

izep (approved):

